PR #16, opened 2019-10-28
CAZ-1234 Get and update payment status

radekmorytko:


src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
```diff
@@ -220,7 +231,7 @@ databaseChangeLog:
                   constraints:
                     nullable: false
   - changeSet:
-      id: 0001.3-1.1
+      id: 0001.4-1.1
```
jaroslawpiotrowski: Shouldn't be `0001.4-1.0`?

src/main/java/uk/gov/caz/psr/dto/GetAndUpdatePaymentStatusResponse.java
```diff
@@ -0,0 +1,36 @@
...
+   */
+  public static GetAndUpdatePaymentStatusResponse from(Payment payment) {
+    return new GetAndUpdatePaymentStatusResponse(payment.getId(),
+        fromInternalStatus(payment.getStatus()));
+  }
+
+  /**
+   * Maps the internal payment's status to the one returned to the REST API caller.
+   */
+  private static PaymentStatus fromInternalStatus(uk.gov.caz.psr.model.PaymentStatus status) {
```
jaroslawpiotrowski: Can you fix import?

src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
radekmorytko: Good catch, fixed.

src/main/java/uk/gov/caz/psr/dto/GetAndUpdatePaymentStatusResponse.java
radekmorytko: I renamed the class so now there are no clashes with names, thanks.

PR #23, opened 2019-10-30
CAZ-1304 Add an integration test for the successful payment journey

radekmorytko:


src/it/resources/application-integration-tests.yml
```diff
@@ -6,3 +6,8 @@ spring:
 #    url: jdbc:postgresql://${postgres.url:localhost:5432}/caz-payments
     hikari:
       maximum-pool-size: 10
+
+services:
+  gov-uk-pay:
+    root-url: http://localhost:1080
+    api-key: P8P5tyxVL7Qb5
```
plenza: This is some dummy key? Maybe we can name it "Dummy" for making it obvious?

radekmorytko: That's right, a random value. Ok, I will change it :)

src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
```diff
@@ -113,21 +114,10 @@ public Payment mapRow(ResultSet resultSet, int i) throws SQLException {
           .externalPaymentId(resultSet.getString("external_payment_id"))
           .status(PaymentStatus.valueOf(resultSet.getString("status")))
           .caseReference(resultSet.getString("case_reference"))
-          .userId(toUuidIfPresent(resultSet, "caz_id"))
```
plenza: I recall that user_id is in database. Why have you removed it here?

radekmorytko: It was 'announced' that for the time being this value is always `null` in the database and it is not used anywhere in the current logic. I think it may cause some confusion, so I decided to remove it for now. I want to restore it once the logic related to fleets are in place.

PR #26, opened 2019-10-31
[CAZ-1354] Updated VehicleEntrance attributes and API specification

musztardem:
* Updated changeSet which alters `DATE_OF_ENTRANCE` attribute to `CAZ_ENTRY_TIMESTAMP`
* Changed `LocalDate dateOfEntrance` to `LocalDateTime dateTimeOfEntrance`
* Updated swagger documentation
* Updated FE-BE contract (_TODO - no edit rights_)

[CAZ_1354 CARD](https://eaflood.atlassian.net/browse/CAZ-1354)

*Note:*
This change will require database drop and recreation on DEV environment

### DIFF
--- a/src/main/java/uk/gov/caz/psr/dto/VehicleEntranceRequest.java
+++ b/src/main/java/uk/gov/caz/psr/dto/VehicleEntranceRequest.java
@@ -1,7 +1,7 @@
 package uk.gov.caz.psr.dto;
 
 import io.swagger.annotations.ApiModelProperty;
-import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.util.UUID;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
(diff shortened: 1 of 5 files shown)


src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
```diff
@@ -259,3 +259,15 @@ databaseChangeLog:
               AFTER INSERT OR UPDATE OR DELETE ON FLEET_VEHICLE
               FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
             stripComments: true
+  - changeSet:
```
plenza: I don't think we have any more envs than DEV, so in that case I would update existing changeset `id: 0001.3-1.0` and drop db (Radek can do it) to reapply

src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
```diff
@@ -36,8 +36,8 @@ databaseChangeLog:
                     nullable: false
                     primaryKeyName: vehicle_entrance_vehicle_entrance_id_pkey
               - column:
-                  name: DATE_OF_ENTRANCE
-                  type: date
+                  name: CAZ_ENTRY_TIMESTAMP
```
radekmorytko: As in https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/27 I needed to change the whole DB schema to be aligned with a new ERD, I also made changes to this column (see: https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/27/files#diff-42afc3d46dbfc50c51fdb4f9f050fb38R39).
I think it will be better to change the schema only in that PR to avoid dropping and creating the database twice - you can safely revert all your changes to `0001-1.0-create_inital_schemas_with_audit_log.yaml`.

PR #30, opened 2019-11-07
[CAZ-1347] Adds Service to connect Payment with existing VehicleEntrants after successful payment

dratka:
Jira Card: https://eaflood.atlassian.net/browse/CAZ-1347

Added new service to inject to successfully finished VehicleEntrantPayments objects vehicleEntratntId if already exists. 
Used in update Payment process.
Added Unit tests
Updated Integration test with a successful path of payment.

### DIFF
--- a/src/it/java/uk/gov/caz/psr/SuccessPaymentsJourneyTestIT.java
+++ b/src/it/java/uk/gov/caz/psr/SuccessPaymentsJourneyTestIT.java
@@ -9,7 +9,7 @@ import com.google.common.base.Joiner;
 import com.google.common.io.Resources;
 import io.restassured.RestAssured;
 import java.time.LocalDate;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
@@ -40,9 +40,11 @@ import uk.gov.caz.psr.model.PaymentStatus;
 import uk.gov.caz.psr.repository.ExternalPaymentsRepository;
 
 @FullyRunningServerIntegrationTest
-@Sql(scripts = "classpath:data/sql/clear-all-payments.sql",
+@Sql(scripts = {"classpath:data/sql/clear-all-payments.sql",
+    "classpath:data/sql/add-vehicle-entrants.sql"},
     executionPhase = ExecutionPhase.BEFORE_TEST_METHOD)
-@Sql(scripts = "classpath:data/sql/clear-all-payments.sql",
+@Sql(scripts = {"classpath:data/sql/clear-all-payments.sql",
+    "classpath:data/sql/clear-all-vehicle-entrants.sql"},
     executionPhase = ExecutionPhase.AFTER_TEST_METHOD)
 public class SuccessPaymentsJourneyTestIT {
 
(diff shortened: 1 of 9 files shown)


src/it/java/uk/gov/caz/psr/SuccessPaymentsJourneyTestIT.java
```diff
@@ -9,12 +9,14 @@
...
 import io.restassured.RestAssured;
 import java.time.LocalDate;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
+import org.apache.tomcat.jni.Local;
```
radekmorytko: Unused import, can be safely removed.

src/it/resources/data/sql/add-vehicle-entrants.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO public.vehicle_entrant(
+  vehicle_entrant_id, caz_entry_timestamp, caz_id, vrn, insert_timestamp)
+  VALUES
+    ('579c6afe-d076-4dd7-a6a4-376991ad4fbf', '2019-11-01 19:10:25-07', 'b8e53786-c5ca-426a-a701-b14ee74857d4', 'ND84VSX', now()),
```
radekmorytko: `insert_timestamp` column has a default value set to `CURRENT_TIMESTAMP`, I think you can omit `now()` and `insert_timestamp`

src/main/java/uk/gov/caz/psr/service/FinalizePaymentService.java
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Adds {@code VehicleEntrantId} to each {@link VehicleEntrantPayment} if {@code Payment} status
+   * if SUCCESS.
+   *
+   * @param payment provided {@link Payment} object
+   * @return {@link Payment} with added VehicleEntrantId if Payment status is SUCCESS
+   */
+  public Payment connectExistingVehicleEntrants(Payment payment) {
+    if (!payment.getStatus().equals(PaymentStatus.SUCCESS)) {
```
radekmorytko: * in Java when comparing enums you can use `==` operator instead of `equals`
* can we add a non-null check precondition here for `payment`?
* if the status is not `success`, I would log the message that the 'connection' will not take place

src/main/java/uk/gov/caz/psr/service/FinalizePaymentService.java
```diff
@@ -0,0 +1,75 @@
...
+
+@Service
+@AllArgsConstructor
+public class FinalizePaymentService {
+
+  private final VehicleEntranceRepository vehicleEntranceRepository;
+
+  /**
+   * Adds {@code VehicleEntrantId} to each {@link VehicleEntrantPayment} if {@code Payment} status
+   * if SUCCESS.
```
radekmorytko: typo: _if SUCCESS_ -> _is SUCCESS_

Maybe can we add the information that *a new* instance of _Payment_ will be created?

PR #31, opened 2019-11-07
[CAZ-1238] VehicleEntranceRepository

musztardem:
**Introduced changes**
* Implemented `public void insertIfNotExist(VehicleEntrance vehicleEntrance)` method in `VehicleEntranceRepository`
* Added specs for preconditions
* Updated db schema with `VehicleEntrant#cazEntryDate`
* Updated associated model, `VehicleEntranceRequest` and swagger docs.

---

**Additional info**

[Link to JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1238)

### DIFF
--- a/src/main/java/uk/gov/caz/psr/model/VehicleEntrance.java
+++ b/src/main/java/uk/gov/caz/psr/model/VehicleEntrance.java
@@ -1,5 +1,6 @@
 package uk.gov.caz.psr.model;
 
+import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.UUID;
 import lombok.Builder;
@@ -14,5 +15,6 @@ public class VehicleEntrance {
   UUID id;
   private final UUID cleanZoneId;
   private final LocalDateTime cazEntryTimestamp;
+  private final LocalDate cazEntryDate;
   private final String vrn;
 }
--- a/src/main/java/uk/gov/caz/psr/repository/VehicleEntranceRepository.java
+++ b/src/main/java/uk/gov/caz/psr/repository/VehicleEntranceRepository.java
@@ -27,13 +27,33 @@ public class VehicleEntranceRepository {
   static final String SELECT_BY_SQL = "SELECT * FROM vehicle_entrant "
       + "WHERE caz_id = ? AND vrn = ? AND to_char(caz_entry_timestamp, 'YYYY-MM-DD') = ?";
 
+  static final String INSERT_IF_NOT_EXISTS_SQL = "INSERT INTO vehicle_entrant "
+      + "(caz_entry_timestamp, caz_entry_date, caz_id, vrn) "
+      + "VALUES (?, ?, ?, ?) "
+      + "ON CONFLICT DO NOTHING";
+
   /**
    * Inserts {@code vehicleEntrance} into database unless it exists.
    *
    * @param vehicleEntrance An entity object which is supposed to be saved in the database.
+   * @throws NullPointerException if {@code vehicleEntrance} is null
+   * @throws IllegalArgumentException if {@link VehicleEntrance#getId()} is not null
    */
   public void insertIfNotExists(VehicleEntrance vehicleEntrance) {
-    // TODO to be implemented in CAZ-1238
+    Preconditions.checkNotNull(vehicleEntrance, "Vehicle Entrance cannot be null");
+    Preconditions.checkArgument(vehicleEntrance.getId() == null,
+            "Vehicle Entrance cannot have ID");
+
+    jdbcTemplate.update(
+        INSERT_IF_NOT_EXISTS_SQL,
+        preparedStatementSetter -> {
+          preparedStatementSetter.setObject(1, vehicleEntrance.getCazEntryTimestamp());
+          preparedStatementSetter.setObject(2,
+              vehicleEntrance.getCazEntryTimestamp().toLocalDate());
+          preparedStatementSetter.setObject(3, vehicleEntrance.getCleanZoneId());
+          preparedStatementSetter.setObject(4, vehicleEntrance.getVrn());
+        }
+    );
   }
 
   /**
--- a/src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
+++ b/src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
@@ -31,6 +31,7 @@ databaseChangeLog:
               - column:
                   name: VEHICLE_ENTRANT_ID
                   type: uuid
+                  defaultValueComputed: uuid_generate_v1mc()
                   constraints:
                     primaryKey: true
                     nullable: false
@@ -51,11 +52,20 @@ databaseChangeLog:
                   constraints:
                     nullable: false
               - column:
+                  name: CAZ_ENTRY_DATE
+                  type: date
+                  constraints:
+                    nullable: false
+              - column:
                   name: INSERT_TIMESTAMP
                   type: timestamp
                   defaultValueComputed: CURRENT_TIMESTAMP
                   constraints:
                     nullable: false
+        - addUniqueConstraint:
+            columnNames: CAZ_ENTRY_DATE,CAZ_ID,VRN
+            constraintName: vehicle_entrant_caz_entry_date_caz_id_vrn_key
+            tableName: VEHICLE_ENTRANT
 #  TODO add unique constraint for (TRAVEL_DATE,CAZ_ID,VRN) when TRAVEL_DATE's presence is confirmed
         - createTable:
             tableName: PAYMENT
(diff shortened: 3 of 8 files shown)


src/main/java/uk/gov/caz/psr/repository/VehicleEntranceRepository.java
```diff
@@ -27,13 +27,32 @@
...
+   * @throws IllegalArgumentException if {@link VehicleEntrance#getId()} is not null
    */
   public void insertIfNotExists(VehicleEntrance vehicleEntrance) {
-    // TODO to be implemented in CAZ-1238
+    Preconditions.checkNotNull(vehicleEntrance, "Vehicle Entrance cannot be null");
+    Preconditions
+      .checkArgument(vehicleEntrance.getId() == null, "Vehicle Entrance cannot have ID");
+
+    jdbcTemplate.update(
+        INSERT_IF_NOT_EXIST_SQL,
```
plenza: Typo in constant name: INSERT_IF_NOT_EXISTS_SQL <- missing S in EXIST(S)

src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
```diff
@@ -50,12 +51,21 @@ databaseChangeLog:
...
+                    nullable: false
               - column:
                   name: INSERT_TIMESTAMP
                   type: timestamp
                   defaultValueComputed: CURRENT_TIMESTAMP
                   constraints:
                     nullable: false
+        - addUniqueConstraint:
+            columnNames: CAZ_ENTRY_DATE,CAZ_ID,VRN
+            constraintName: caz_entry_date_caz_id_vrn_key
```
plenza: If possible please change "key" to "unique_constraint"

src/main/java/uk/gov/caz/psr/model/VehicleEntrance.java
```diff
@@ -9,10 +10,11 @@
  * An entity which represents a row stored in the database in {@code VEHICLE_ENTRANCE} table.
  */
 @Data
-@Builder
+@Builder(toBuilder = true)
 public class VehicleEntrance {
```
plenza: Table name is called "VEHICLE_ENTRANT" but all models and Dtos are "Entrance" - maybe make them uniform?

Note: I'm not 100% sure we should do this thou... Entrance sounds better in my head. Maybe other guys can vote for this?

src/test/java/uk/gov/caz/psr/controller/VehicleEntranceControllerTest.java
```diff
@@ -83,9 +83,25 @@ public void shouldReturn400StatusCodeWhenDateOfEntranceIsNull() throws Exception
...
+        String payload = toJson(new VehicleEntranceRequest(ANY_CLEAN_ZONE_ID, todayDateTime(), null, ANY_VALID_VRN));
+
+        mockMvc.perform(post(PATH)
+          .content(payload)
+          .header(Constants.X_CORRELATION_ID_HEADER, ANY_CORRELATION_ID)
+          .contentType(MediaType.APPLICATION_JSON_UTF8_VALUE)
+          .accept(MediaType.APPLICATION_JSON_UTF8_VALUE))
+          .andExpect(status().isBadRequest())
+          .andExpect(header().string(Constants.X_CORRELATION_ID_HEADER, ANY_CORRELATION_ID));
+      // TODO add assertions for a message
```
plenza: Any chance to fix this TODO?

src/test/java/uk/gov/caz/psr/controller/VehicleEntranceControllerTest.java
```diff
@@ -137,7 +153,8 @@ private String toJson(VehicleEntranceRequest request) {
     return objectMapper.writeValueAsString(request);
   }
 
-  private LocalDateTime today() {
+  private LocalDateTime todayDateTime() {
     return LocalDateTime.now();
   }
+  private LocalDate todayDate() { return LocalDate.now(); }
```
plenza: Please add new line above and format this simple helper to our standard format of:
```
private LocalDate todayDate() {
  return LocalDate.now();
}
```

src/test/java/uk/gov/caz/psr/service/VehicleEntranceServiceTest.java
```diff
@@ -63,6 +65,7 @@ public void shouldCallRepositoryWhenVehicleEntranceIsValid() {
     VehicleEntrance vehicleEntrance = VehicleEntrance.builder()
```
plenza: In this test, we don't care about details of input `vehicleEntrance` object, but only about the fact that proper repository method was invoked, with correct object. My suggestion is to create some sample `VehicleEntrance` object in `TestObjects` class (static objects). Object can be named `SAMPLE_ENTRANCE` or `ANY_ENTRANCE` - and use it here instead of creating new object.

src/main/java/uk/gov/caz/psr/model/VehicleEntrance.java
```diff
@@ -9,10 +10,11 @@
  * An entity which represents a row stored in the database in {@code VEHICLE_ENTRANCE} table.
  */
 @Data
-@Builder
+@Builder(toBuilder = true)
```
plenza: `toBuilder = true`  according to Lombok doc:
If using @Builder to generate builders to produce instances of your own class (this is always the case unless adding @Builder to a method that doesn't return your own type), you can use @Builder(toBuilder = true) to also generate an instance method in your class called toBuilder(); it creates a new builder that starts out with all the values of this instance.

I was not able to find any place where you used this instance builder. I could have missed it however.

src/main/java/uk/gov/caz/psr/model/VehicleEntrance.java
dratka: I'm working on it on my branch to update all the places in the code from entrance to entrant. so I'll merge it with my changes when it is ready.

src/main/java/uk/gov/caz/psr/dto/VehicleEntranceRequest.java
```diff
@@ -22,6 +23,10 @@
   @NotNull
   LocalDateTime cazEntryTimestamp;
 
+  @ApiModelProperty(value = "${swagger.model.descriptions.vehicle-entrance.caz-entry-date}")
+  @NotNull
+  LocalDate cazEntryDate;
```
radekmorytko: I don't think we should change the API. We can (and should, in my opinion) extract the _date_ part from the `cazEntryTimestamp`

src/main/java/uk/gov/caz/psr/model/VehicleEntrance.java
radekmorytko: _Entrance_ also sounds better for me, but the change is suggested from our colleagues from the UK and is approved from JAQU. Therefore we decided to stick to it.

src/main/java/uk/gov/caz/psr/repository/VehicleEntranceRepository.java
```diff
@@ -27,13 +27,32 @@
...
    * Inserts {@code vehicleEntrance} into database unless it exists.
    *
    * @param vehicleEntrance An entity object which is supposed to be saved in the database.
+   * @throws NullPointerException if {@code vehicleEntrance} is null
+   * @throws IllegalArgumentException if {@link VehicleEntrance#getId()} is not null
    */
   public void insertIfNotExists(VehicleEntrance vehicleEntrance) {
-    // TODO to be implemented in CAZ-1238
+    Preconditions.checkNotNull(vehicleEntrance, "Vehicle Entrance cannot be null");
+    Preconditions
```
radekmorytko: Minor - can we slightly change the formatting here and put a new line after the comma?

src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
radekmorytko: I think Antek used the same naming convention which I (wrongly! I missed the _tablename_ part) used before for a unique constraint. The convention comes from stackoverflow:
> `{tablename}_{columnname(s)}_{suffix}`
> where the suffix is one of the following:
> ...
> `key` for a Unique constraint;

so in our case, if we want to stick to it, it should be: `vehicle_entrant_caz_entry_date_caz_id_vrn_key`

source: https://stackoverflow.com/a/4108266

src/test/java/uk/gov/caz/psr/controller/VehicleEntranceControllerTest.java
radekmorytko: I will do it in a separate commit - currently our global exception handler kicks in and all validation error messages are ignored in the response body.

src/main/resources/db/changelog/changes/0001-1.0-create_inital_schemas_with_audit_log.yaml
plenza: Sounds good, keep it

PR #37, opened 2019-11-12
Send payment receipt

kingarj:


pom.xml
```diff
@@ -1,3 +1,4 @@
+
```
Thomas-95IS: Whitespace.

src/main/java/uk/gov/caz/psr/Application.java
```diff
@@ -3,12 +3,9 @@
...
 
-@SpringBootApplication(scanBasePackages = {
-    "uk.gov.caz.psr.configuration",
-    "uk.gov.caz.psr.controller",
-    "uk.gov.caz.psr.repository",
-    "uk.gov.caz.psr.service"
-})
+@SpringBootApplication(scanBasePackages = {"uk.gov.caz.psr.configuration",
+    "uk.gov.caz.psr.controller", "uk.gov.caz.psr.messaging",
+    "uk.gov.caz.psr.repository", "uk.gov.caz.psr.service"})
```
Thomas-95IS: Formatting change looks a little uglier, to my humble mind.

src/main/java/uk/gov/caz/psr/configuration/SpringCloudAwsConfig.java
```diff
@@ -0,0 +1,83 @@
...
+ * AWS client builder configuration.
+ */
+@Configuration
+@Slf4j
+public class SpringCloudAwsConfig {
+
+  @Value("${cloud.aws.region.static}")
+  private String region;
+
+  public static final String DUMMY = "dummy";
```
Thomas-95IS: Irrelevant variable assignment?

src/main/java/uk/gov/caz/psr/configuration/SpringCloudAwsConfig.java
```diff
@@ -0,0 +1,83 @@
...
+
+    if (Strings.isNullOrEmpty(sqsEndpoint)) {
+      throw new IllegalStateException(
+          "SQS endpoint must be overridden when running with "
+              + "Localstack! Please set in 'aws.sqs.endpoint' property");
+    }
+
+    log.info("Using '{}' as SQS Endpoint", sqsEndpoint);
+
+
```
Thomas-95IS: Whitespace.

src/main/java/uk/gov/caz/psr/configuration/SpringCloudAwsConfig.java
```diff
@@ -0,0 +1,83 @@
...
+  @Primary
+  @Profile({"integration-tests", "localstack"})
+  public AmazonSQSAsync sqsLocalstackClient(
+      @Value("${aws.sqs.endpoint:}") String sqsEndpoint) {
+    log.info("Running Spring-Boot app locally using Localstack. ");
+
+    if (Strings.isNullOrEmpty(sqsEndpoint)) {
+      throw new IllegalStateException(
+          "SQS endpoint must be overridden when running with "
+              + "Localstack! Please set in 'aws.sqs.endpoint' property");
```
Thomas-95IS: Why wouldn't we provide this as a default in `application.yml`? Localstack endpoints are standardized.

src/main/java/uk/gov/caz/psr/messaging/MessagingClient.java
```diff
@@ -0,0 +1,50 @@
...
+   */
+  public void publishMessage(SendEmailRequest message) {
+    log.info("Publishing message with reference number: {}",
+        message.getReference());
+
+    Map<String, Object> headers = new HashMap<String, Object>();
+    headers.put(SqsMessageHeaders.SQS_GROUP_ID_HEADER, messageGroupId);
+    headers.put(SqsMessageHeaders.SQS_DEDUPLICATION_ID_HEADER,
+        UUID.randomUUID().toString());
+    // headers.put("contentType", "application/json");
```
Thomas-95IS: Leftover comment.

src/main/java/uk/gov/caz/psr/service/GetAndUpdatePaymentsService.java
```diff
@@ -21,17 +22,19 @@
...
+   * database or it has {@code null} external identifier.
    *
    * @throws NullPointerException if {@code id} is null
    */
   public Optional<Payment> getExternalPaymentAndUpdateStatus(UUID id) {
     Preconditions.checkNotNull(id, "ID cannot be null");
 
-    Payment internalPayment = internalPaymentsRepository.findById(id).orElse(null);
+    Payment internalPayment =
+        internalPaymentsRepository.findById(id).orElse(null);
```
Thomas-95IS: If dealing with `Optionals`, it might be better to return an `Optional` and then call `Optional.isPresent()` rather than casting to a null value.

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
```diff
@@ -1,100 +1,177 @@
...
   private final FinalizePaymentService finalizePaymentService;
   private final PaymentRepository internalPaymentsRepository;
+  private final MessagingClient messagingClient;
+
+  @Value("${services.sqs.template-id}")
+  String templateId;
+
+  /**
+   * Constructor.
+   */
```
Thomas-95IS: More insightful javadoc?

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
```diff
@@ -1,100 +1,177 @@
...
+    Map<String, String> personalisationMap = new HashMap<String, String>();
+    personalisationMap.put("amount", Integer.toString(amount));
+
+    String personalisation;
+    try {
+      personalisation = objectMapper.writeValueAsString(personalisationMap);
+    } catch (JsonProcessingException e) {
+      log.error(e.getMessage());
+      personalisation = "{}";
+    }
```
Thomas-95IS: Are we happy with this logic? If can't process JSON, still attempt email sending? Sending an email with no personalisation might be quite confusing for a user.

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
```diff
@@ -1,100 +1,177 @@
...
+   *                         entrants' ids set (if applicable).
    */
-  private Payment finalizePaymentProcessing(Payment internalPayment, Payment externalPayment) {
-    if (internalPayment.getExternalPaymentStatus() == externalPayment.getExternalPaymentStatus()) {
-      log.info("External payment status is the same as the one from database and equal to '{}', "
-          + "skipping updating the database", internalPayment.getExternalPaymentStatus());
+  private Payment finalizePaymentProcessing(Payment internalPayment,
+      Payment externalPayment) {
+    if (internalPayment.getExternalPaymentStatus() == externalPayment
+        .getExternalPaymentStatus()) {
```
Thomas-95IS: `.equals()` would probably be safer here.

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
```diff
@@ -1,100 +1,177 @@
...
+      log.info(
+          "External payment status is the same as the one from database and equal to '{}', "
+              + "skipping updating the database",
+          internalPayment.getExternalPaymentStatus());
```
Thomas-95IS: Could we include an ID for the payment here, so that we can track it through the logs?

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
```diff
@@ -1,100 +1,177 @@
...
       return externalPayment;
     }
-    log.info("Found the external payment, updating its status to '{}' in the database, "
-            + "current status: '{}'", externalPayment.getExternalPaymentStatus(),
+    log.info(
+        "Found the external payment, updating its status to '{}' in the database, "
+            + "current status: '{}'",
+        externalPayment.getExternalPaymentStatus(),
```
Thomas-95IS: Could we include an ID for the payment here, so that we can track it through the logs?

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
Thomas-95IS: Or Notify will throw a 400.

src/main/java/uk/gov/caz/psr/Application.java
kingarj: @Thomas-95IS I agree, unfortunately my formatter seems to think this is better.

src/main/java/uk/gov/caz/psr/configuration/SpringCloudAwsConfig.java
kingarj: We don't use a specific endpoint in the regular configuration, just a region. Then we let the library figure out how to talk to SQS using the AWS credentials in our profile.

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
kingarj: Good point. The only possible scenario this happens is if the integer is somehow invalid when it is written into the HashMap (which is unlikely in itself as this is a value we get from Gov.UK Pay). I have made the handling of this exception the responsibility of the parent method and logged a warning rather than attempting to send the email anyway.

radekmorytko:
In my opinion now `UpdatePaymentWithExternalDataService` has way too many responsibilities (it qualified for refactoring before - I'm working on it now, but will wait for your changes to be merged to develop to avoid conflicts): it gets the external status, updates it in the database alongside with associating `vehicle_entrant` records with `vehicle_entrant_payment` records, and now sending emails.

Sending an email is a *side effect* of a successful update operation with `ExternalPaymentStatus.SUCCESS` status. For processing side effects I would recommend employing spring events mechanism. Suggested approach:
- once the payment has been successfully updated, we can send an event. This should be done just before the value is returned from `uk.gov.caz.psr.service.GetAndUpdatePaymentsService#getExternalPaymentAndUpdateStatus` method
- for sending an event you can use `ApplicationEventPublisher` bean (can be injected)
- an event should extend `ApplicationEvent` and can contain e.g. the whole `Payment` object
- create a listener bean; the handling method can filter the events which contain only 'SUCCESS' external status
- the listener bean either prepares the message (`SendEmailRequest` class) or delegates it to another class (I think the latter is a better approach); once `SendEmailRequest` is ready, it is sent using the injected `MessagingClient`. Alternatively, the whole logic is in another class and listener only 'catches' the event and calls that class for further processing
- I can see now that `SendEmailRequest` contains `templateId`, which comes from property - then maybe it is a good idea to create a separate class which creates instances of this class and accepts `Payment`s?
- you can add `email` field to `Payment` class and comment it that this is a transient field
- currently when someone makes a request to `/v1/payments/{id}` and the external status is `SUCCESS`, the email is always sent. I'm not sure if this is a desired behaviour

Let me know what you think, I won't review formatting, whitespaces and so on as at this stage I think they're of less relevance.

src/main/java/uk/gov/caz/psr/service/UpdatePaymentWithExternalDataService.java
plenza: You can annotate this class with `@AllArgsConstructor` and delete this constructor.

src/main/java/uk/gov/caz/psr/messaging/MessagingClient.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.cloud.aws.messaging.core.QueueMessagingTemplate;
+import org.springframework.cloud.aws.messaging.core.SqsMessageHeaders;
+import org.springframework.stereotype.Component;
+import uk.gov.caz.psr.dto.SendEmailRequest;
+
+@Component
+@Slf4j
+public class MessagingClient {
```
plenza: Unfortunately we need to put Javadocs: before class, methods (private, protected and public) and properties.

src/main/java/uk/gov/caz/psr/Application.java
plenza: We obviously have different formatting configuration - it is a matter of finding proper setting - unfortunately we don't use Eclipse so I can't help here, but I'm sure it is there somewhere....

src/main/java/uk/gov/caz/psr/configuration/SpringCloudAwsConfig.java
```diff
@@ -0,0 +1,83 @@
...
+  @Value("${cloud.aws.region.static}")
+  private String region;
+
+  public static final String DUMMY = "dummy";
+
+  /**
+   * Returns an instance of {@link AmazonSQSAsync} which is used to send a
+   * message to a SQS queue mocked by Localstack.
+   *
+   * @param  sqsEndpoint           An endpoint of mocked SQS. Cannot be empty or
```
plenza: This is another formatting issue - there shouldn't be right aligning to the longest param name, just 1 space and description. Actually CheckStyle should validate it and break the build... no idea why it didn't...

src/main/java/uk/gov/caz/psr/dto/SendEmailRequest.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.caz.psr.dto;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class SendEmailRequest {
```
plenza: Javadocs

plenza:
> * for sending an event you can use `ApplicationEventPublisher` bean (can be injected)
> * an event should extend `ApplicationEvent` and can contain e.g. the whole `Payment` object
> * create a listener bean; the handling method can filter the events which contain only 'SUCCESS' external status
> * the listener bean either prepares the message (`SendEmailRequest` class) or delegates it to another class (I think the latter is a better approach); once `SendEmailRequest` is ready, it is sent using the injected `MessagingClient`. Alternatively, the whole logic is in another class and listener only 'catches' the event and calls that class for further processing

My 2 cents: just make sure that event raising and handling is fully synchronous, because on Lambdas, AWS will not wait for any background threads to finish - when main thread is done, all JVM is halted.

src/main/java/uk/gov/caz/psr/Application.java
radekmorytko: Can you try using the [one](https://github.com/InformedSolutions/JAQU-CAZ-National-Taxi-Register/blob/develop/developer-resources/eclipse-java-google-style.xml) from NTR?

radekmorytko:
> > * for sending an event you can use `ApplicationEventPublisher` bean (can be injected)
> > * an event should extend `ApplicationEvent` and can contain e.g. the whole `Payment` object
> > * create a listener bean; the handling method can filter the events which contain only 'SUCCESS' external status
> > * the listener bean either prepares the message (`SendEmailRequest` class) or delegates it to another class (I think the latter is a better approach); once `SendEmailRequest` is ready, it is sent using the injected `MessagingClient`. Alternatively, the whole logic is in another class and listener only 'catches' the event and calls that class for further processing
> 
> My 2 cents: just make sure that event raising and handling is fully synchronous, because on Lambdas, AWS will not wait for any background threads to finish - when main thread is done, whole JVM including all active threads is halted.

I have checked it and by default Spring make synchronous calls. If we want to make async ones, we need to use the `@Async` annotation. More can  be found [here](https://docs.spring.io/autorepo/docs/spring-framework/5.1.7.RELEASE/spring-framework-reference/core.html#context-functionality-events-async)

src/main/java/uk/gov/caz/psr/Application.java
kingarj: Thanks @radekmorytko, I have been using the Google style Eclipse formatter. If the preference is for a line break after each then I can probably find a way of turning it off so I can commit this specific change.

kingarj:
@radekmorytko @plenza Thank you both for your feedback on this. I'm going to take it away and refactor as suggested.

src/main/java/uk/gov/caz/psr/configuration/SpringCloudAwsConfig.java
```diff
@@ -0,0 +1,79 @@
...
+
+    if (Strings.isNullOrEmpty(sqsEndpoint)) {
+      throw new IllegalStateException(
+          "SQS endpoint must be overridden when running with "
+              + "Localstack! Please set in 'aws.sqs.endpoint' property");
+    }
+
+    log.info("Using '{}' as SQS Endpoint", sqsEndpoint);
+
+    AmazonSQSAsyncClientBuilder builder =
```
radekmorytko: When running with this configuration locally integration tests failed:
```
[ERROR] successPaymentsJourney  Time elapsed: 6.434 s  <<< ERROR!
com.amazonaws.SdkClientException: Unable to load AWS credentials from any provider in the chain: [EnvironmentVariableCredentialsProvider: Unable to load AWS credentials from environment variables (AWS_ACCESS_KEY_ID (or AWS_ACCESS_KEY) and AWS_SECRET_KEY (or AWS_SECRET_ACCESS_KEY)), SystemPropertiesCredentialsProvider: Unable to load AWS credentials from Java system properties (aws.accessKeyId and aws.secretKey), com.amazonaws.auth.profile.ProfileCredentialsProvider@2f2464ae: No AWS profile named 'default', com.amazonaws.auth.EC2ContainerCredentialsProviderWrapper@38dfff: Unable to load credentials from service endpoint]
        at uk.gov.caz.psr.SuccessPaymentsJourneyTestIT.startMockServer(SuccessPaymentsJourneyTestIT.java:89)
```
The solution is to set a dummy credentials provider, e.g. (additionally I chained invocations on `builder`)
```java
  return AmazonSQSAsyncClientBuilder.standard()
      .withCredentials(dummyCredentialsProvider())
      .withEndpointConfiguration(new EndpointConfiguration(sqsEndpoint, region))
      .build();
}

private AWSStaticCredentialsProvider dummyCredentialsProvider() {
  return new AWSStaticCredentialsProvider(
      new BasicAWSCredentials("dummy-access-key", "dummy-secret-key")
  );
}
```

src/main/java/uk/gov/caz/psr/model/events/SendEmailEvent.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.context.ApplicationEvent;
+import uk.gov.caz.psr.model.Payment;
+
+/**
+ * Event raised to enable sending of receipt email.
+ */
+@Getter
+public class SendEmailEvent extends ApplicationEvent {
+  private static final long serialVersionUID = 1911485375258518221L;
+  private Payment payment;
```
radekmorytko: Can be `final`

src/main/java/uk/gov/caz/psr/service/PaymentReceiptService.java
```diff
@@ -0,0 +1,41 @@
...
+   * 
+   * @param email the recipient of the email
+   * @param amount the total cost of their CAZ charge
+   * @return SendEmailRequest
+   * @throws JsonProcessingException if the amount cannot be serialized into a json string
+   */
+  public SendEmailRequest buildSendEmailRequest(String email, int amount)
+      throws JsonProcessingException {
+    ObjectMapper objectMapper = new ObjectMapper();
+    Map<String, String> personalisationMap = new HashMap<String, String>();
```
radekmorytko: Can we move creating the personalisation map and converting it to JSON to a separate method? Additionally, we can make use of `java.util.Collections#singletonMap` method:
```java 
public SendEmailRequest buildSendEmailRequest(String email, int amount)
    throws JsonProcessingException {
  return SendEmailRequest.builder()
      .emailAddress(email)
      .personalisation(createPersonalisationPayload(amount))
      .templateId(templateId)
      .build();
}

private String createPersonalisationPayload(int amount) throws JsonProcessingException {
  Map<String, String> personalisationMap = Collections.singletonMap("amount",
      Integer.toString(amount));
  return objectMapper.writeValueAsString(personalisationMap);
}
```

src/main/java/uk/gov/caz/psr/model/events/SendEmailEvent.java
```diff
@@ -0,0 +1,20 @@
...
+
+import lombok.Getter;
+import org.springframework.context.ApplicationEvent;
+import uk.gov.caz.psr.model.Payment;
+
+/**
+ * Event raised to enable sending of receipt email.
+ */
+@Getter
+public class SendEmailEvent extends ApplicationEvent {
```
radekmorytko: Minor: my suggestion is to rename this class to e.g. `PaymentStatusUpdatedEvent`. Why? The code in `uk.gov.caz.psr.service.UpdatePaymentWithExternalDataService#finalizePaymentProcessing` raises this event and doesn't really care what will be done with (if there are _any_ listeners at all).
However, we create a dedicated listener which knows exactly what needs to be done. And here comes another renaming suggestions:
-  `EmailEventListener` to `PaymentReceiptSender`
- `onApplicationEvent` to `onPaymentStatusUpdated`

src/main/java/uk/gov/caz/psr/service/listener/EmailEventListener.java
```diff
@@ -0,0 +1,49 @@
...
+import uk.gov.caz.psr.service.PaymentReceiptService;
+
+/**
+ * Listener to pick up email events and use messaging client to build and send an email.
+ *
+ */
+@Component
+@Slf4j
+@RequiredArgsConstructor
+public class EmailEventListener implements ApplicationListener<SendEmailEvent> {
```
radekmorytko: There is no need to implement this interface if you annotate `onApplicationEvent` method with `@EventListener` annotation.

src/main/java/uk/gov/caz/psr/service/listener/EmailEventListener.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private final MessagingClient messagingClient;
+  private final PaymentReceiptService paymentReceiptService;
+
+  @Override
+  public void onApplicationEvent(SendEmailEvent event) {
+    Payment payment = event.getPayment();
+
+    if (payment.getExternalPaymentStatus().equals(ExternalPaymentStatus.SUCCESS)) {
```
radekmorytko: If we use `@EventListener` annotation for this method, we can move this check to the annotation itself:
```java
@EventListener(condition = "#event.payment.externalPaymentStatus.name() == 'SUCCESS'")
public void onApplicationEvent(PaymentStatusUpdatedEvent event) {
...
```

pom.xml
```diff
@@ -115,7 +115,18 @@
       <groupId>com.amazonaws.secretsmanager</groupId>
       <artifactId>aws-secretsmanager-jdbc</artifactId>
       <version>1.0.2</version>
-    </dependency>
+    </dependency>    
+    <dependency>
+	    <groupId>com.amazonaws</groupId>
+	    <artifactId>amazon-sqs-java-messaging-lib</artifactId>
+	    <version>1.0.8</version>
+	    <type>jar</type>
```
radekmorytko: `type` by default is equal to `jar`, you can omit it here.

src/main/java/uk/gov/caz/psr/messaging/MessagingClient.java
```diff
@@ -0,0 +1,52 @@
...
+ * A wrapper class for an external queuing system with the ability to publish messages to the queue.
+ *
+ */
+@Component
+@Slf4j
+public class MessagingClient {
+
+  private final QueueMessagingTemplate messagingTemplate;
+
+  @Value("${services.sqs.new-queue-name}")
```
radekmorytko: Please see my comment: https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/37/files#r346719271

src/main/java/uk/gov/caz/psr/messaging/MessagingClient.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * A method to publish a message to a queue.
+   * 
+   * @param message the message to be published
+   */
+  public void publishMessage(SendEmailRequest message) {
+    log.info("Publishing message with reference number: {}", message.getReference());
+
+    Map<String, Object> headers = new HashMap<String, Object>();
```
radekmorytko: Maybe extract header creation to a separate method? For example:
```java
  private Map<String, Object> createHeaders() {
    return ImmutableMap.of(
        SqsMessageHeaders.SQS_GROUP_ID_HEADER, messageGroupId,
        SqsMessageHeaders.SQS_DEDUPLICATION_ID_HEADER, UUID.randomUUID().toString(),
        "contentType", "application/json"
    );
  
```

src/main/java/uk/gov/caz/psr/messaging/MessagingClient.java
```diff
@@ -0,0 +1,52 @@
...
+    headers.put(SqsMessageHeaders.SQS_GROUP_ID_HEADER, messageGroupId);
+    headers.put(SqsMessageHeaders.SQS_DEDUPLICATION_ID_HEADER, UUID.randomUUID().toString());
+    headers.put("contentType", "application/json");
+
+    log.info("Queue is: {}", newQueue);
+    log.info("Message is: {}", message.toString());
```
radekmorytko: There is no need to call `toString` method explicitly - those objects will be converted to string implicitly.

PR #39, opened 2019-11-13
[CAZ-1429] Updated PUT /v1/charge-settlement/payment-status endpoint based on Open API

dratka:
Jira Card: https://eaflood.atlassian.net/browse/CAZ-1429
Updated the controller action based on OpenAPI
Added Basic Tests

### DIFF
--- a/src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
+++ b/src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
@@ -1,6 +1,7 @@
 package uk.gov.caz.psr.dto;
 
 import io.swagger.annotations.ApiModelProperty;
+import javax.validation.constraints.Max;
 import javax.validation.constraints.NotNull;
 import lombok.Builder;
 import lombok.Value;
@@ -13,6 +14,7 @@ public class PaymentStatusResponse {
   ChargeSettlementPaymentStatus chargeSettlementPaymentStatus;
 
   @ApiModelProperty(value = "${swagger.model.descriptions.payment-status.payment-id}")
+  @Max(255)
   String paymentId;
 
   @ApiModelProperty(value = "${swagger.model.descriptions.payment-status.case-reference}")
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
```diff
@@ -2,14 +2,17 @@
...
 import lombok.Value;
 
 @Value
+@Builder
 public class PaymentStatusResponse {
   @ApiModelProperty(value = "${swagger.model.descriptions.payment-status.payment-status}")
   @NotNull
   ChargeSettlementPaymentStatus chargeSettlementPaymentStatus;
 
+  @NotNull
```
radekmorytko: I'm not sure whether we want this field to be non-nullable - according to ICD it is optional (or you thought of sending `""` as the value is acceptable, but this might be confusing). And being optional makes sense, because not all payments are done via _gov uk pay_, for the ones done in LAs this value is not set.
My suggestion here, though, is to set some maximum length constraint, e.g. `255` (the current db constraint)

src/main/java/uk/gov/caz/psr/dto/PaymentStatusUpdateDetails.java
```diff
@@ -0,0 +1,35 @@
...
+  @ApiModelProperty(value = "${swagger.model.descriptions.payment-status-update.date-caz-entry}")
+  @NotNull
+  LocalDate dateOfCazEntry;
+
+  @ApiModelProperty(value = "${swagger.model.descriptions.payment-status-update.payment-status}")
+  @NotNull
+  ChargeSettlementPaymentStatus chargeSettlementPaymentStatus;
+
+  @ApiModelProperty(value = "${swagger.model.descriptions.payment-status-update.payment-id}")
+  @NotNull
```
radekmorytko: See my comment above: https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/39/files#r346686997

PR #40, opened 2019-11-13
[CAZ-1416] GET ChargeSettlement endpoint with fake data

musztardem:
**Introduced changes**
* Implemented `getPaymentStatus` endpoint which responses with some stubbed data
* Added `PaymentStatusRequest` class to represent params
* Added specs for the implemented endpoint

**Additional info**

[Link to JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1416)

### DIFF
--- a/src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
+++ b/src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
@@ -3,7 +3,9 @@ package uk.gov.caz.psr.controller;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
+import uk.gov.caz.psr.dto.ChargeSettlementPaymentStatus;
 import uk.gov.caz.psr.dto.PaymentInfoResponse;
+import uk.gov.caz.psr.dto.PaymentStatusRequest;
 import uk.gov.caz.psr.dto.PaymentStatusResponse;
 import uk.gov.caz.psr.dto.PaymentStatusUpdateRequest;
 import uk.gov.caz.psr.dto.PaymentUpdateSuccessResponse;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
```diff
@@ -25,8 +28,23 @@
...
-  public ResponseEntity<PaymentStatusResponse> getPaymentStatus() {
-    throw new UnsupportedOperationException();
+  public ResponseEntity<PaymentStatusResponse> getPaymentStatus(PaymentStatusRequest request) {
+    /* TODO: Fetch:
+     *    paymentStatus from VEHICLE_ENTRANT_PAYMENT table
+     *    paymentId from PAYMENT table
+     *    paymentStatus from VEHICLE_ENTRANT_PAYMENT table
+     *
+     **/
+    log.info("Received VRN: " + request.getVrn());
```
plenza: `PaymentStatusRequest` is `@Value` - it means it will be enriched with proper `toString()` method. All in all you can log whole object here and get nice detailed output with property names and values:
```
log.info("Received request: {}", request);
```

src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
```diff
@@ -25,8 +28,23 @@
...
+    log.info("Received date of entry: " + request.getDateOfCazEntry().toString());
+
+    PaymentStatusResponse fakeResponse = new PaymentStatusResponse(
```
plenza: Would probably benefit from `@Builder` on `PaymentStatusResponse` class - currently, constructor takes 3 parameters, where last 2 are "enigmatic" - some strings... with builder it will be obvious what is being constructed and with what values

src/test/java/uk/gov/caz/psr/controller/ChargeSettlementControllerTest.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final String ANY_VALID_DATE_STRING = LocalDate.now().toString();
+  private static final String ANY_CORRELATION_ID = "03d339e2-875f-4b3f-9dfa-1f6aa57cc119";
+  private static final String PAYMENT_STATUS_PATH = ChargeSettlementController.BASE_PATH
+      + "/" + ChargeSettlementController.PAYMENT_STATUS_PATH;
+
+  @Nested
+  class PaymentStatus {
+    @Test
+    public void shouldReturn400StatusCodeWhenVrnIsMissing() throws Exception {
+      mockMvc.perform(get(PAYMENT_STATUS_PATH)
```
plenza: Do not think too much about errors and validations - there is separate task for that in the next sprint

src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
```diff
@@ -25,8 +28,22 @@
...
+     **/
+    log.info("Received request: {}", request);
+
+    PaymentStatusResponse fakeResponse = PaymentStatusResponse.builder()
+        .chargeSettlementPaymentStatus(ChargeSettlementPaymentStatus.PAID)
+        .paymentId("350be6da-10f1-41fe-9840-98c738ec763e")
+        .caseReference("sample-case-reference")
+        .build();
+
+    return ResponseEntity.status(HttpStatus.OK).body(fakeResponse);
```
radekmorytko: Minor: can be changed to
```
return ResponseEntity.ok(fakeResponse)
```

PR #41, opened 2019-11-14
CAZ-1345 Additionally update data in vehicle_entrant and vehicle_entr…

radekmorytko:
…ant_payment tables when processing a dangling payment


src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
```diff
@@ -0,0 +1,63 @@
...
+      ExternalPaymentStatus externalStatus = externalStatusFetcher
+          .fetchExternalStatusFor(danglingPayment);
+
+      if (externalStatus == danglingPayment.getExternalPaymentStatus()) {
+        log.info("The status of a dangling payment has not changed and is equal to '{}', "
+            + "aborting the update", danglingPayment.getExternalPaymentStatus());
+        return;
+      }
+
+      paymentStatusUpdater.updateWithStatus(loadVehicleEntrantPayments(
```
plenza: Please add log in which you'll state to what status you are changing

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
```diff
@@ -0,0 +1,63 @@
...
+          danglingPayment), externalStatus);
+    } finally {
+      long elapsed = stopwatch.elapsed(TimeUnit.MILLISECONDS);
+      log.info("Processing the dangling payment '{}' took {}ms", danglingPayment.getId(), elapsed);
```
plenza: Thank you

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  /**
+   * For the passed {@code payment} it loads all associated vehicle entrant records. A new
+   * {@link Payment} instance is created with the newly fetched records ann all previous
```
plenza: typo ann -> and

src/main/java/uk/gov/caz/psr/service/GetAndUpdatePaymentsService.java
```diff
@@ -31,21 +33,28 @@
   public Optional<Payment> getExternalPaymentAndUpdateStatus(UUID id) {
     Preconditions.checkNotNull(id, "ID cannot be null");
 
-    Payment internalPayment = internalPaymentsRepository.findById(id).orElse(null);
+    Payment payment = internalPaymentsRepository.findById(id).orElse(null);
```
plenza: Sorry for perhaps silly question: how is it possible that we get ID and do not find any payment in database? Is this valid use case from user perspective? My understanding is that the only way for this to happen is when frontend passes invalid UUID, right? Should we propagate this 'error' by returning Optional? I would expect to fail fast, throw exception here and return 400 to the caller.... Or am I missing something (most probably)?

src/main/java/uk/gov/caz/psr/service/PaymentExternalStatusFetcher.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import uk.gov.caz.psr.dto.external.GetPaymentResult;
+import uk.gov.caz.psr.model.ExternalPaymentStatus;
+import uk.gov.caz.psr.model.Payment;
+import uk.gov.caz.psr.repository.ExternalPaymentsRepository;
+
+/**
+ * A class which is responsible for getting an external status of a given payment, updating it in
```
plenza: Class description here is not reflecting what the code does. Leftover from copy-paste?

src/main/java/uk/gov/caz/psr/service/PaymentExternalStatusFetcher.java
```diff
@@ -0,0 +1,41 @@
...
+@Slf4j
+@AllArgsConstructor
+public class PaymentExternalStatusFetcher {
+
+  private final ExternalPaymentsRepository externalPaymentsRepository;
+
+  /**
+   * Gets an external status of {@code payment}.
+   *
+   * @throws NullPointerException if {@code payment} is null
```
plenza: missing `@return` section

src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private final PaymentWithExternalStatusBuilder paymentWithExternalStatusBuilder;
+  private final PaymentRepository internalPaymentsRepository;
+  private final TransientVehicleEntrantsLinker transientVehicleEntrantsLinker;
+
+  /**
+   * Updates {@code payment} with a new {@code status}.
+   */
+  public Payment updateWithStatus(Payment payment, ExternalPaymentStatus status) {
+    Preconditions.checkNotNull(payment, "Payment cannot be null");
```
plenza: This concerns all your code: Would it be possible to wrap preconditions inside helper private method and maybe call it `checkPreconditions(params)`? They can get long and they obscure real business functionality of the method. And what and how they validate is in most cases not important to the reader.

src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+        payment, status);
+
+    Payment updatedPayment = transientVehicleEntrantsLinker.associateExistingVehicleEntrantsWith(
+        paymentWithNewStatus);
+
+    internalPaymentsRepository.update(updatedPayment);
+
+    return updatedPayment;
+  }
+
```
plenza: drop me please

src/main/java/uk/gov/caz/psr/service/PaymentWithExternalStatusBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+    Preconditions.checkNotNull(payment, "Payment cannot be null");
+    Preconditions.checkNotNull(newStatus, "newStatus cannot be null");
+    Preconditions.checkArgument(newStatus != payment.getExternalPaymentStatus(),
+        "Status cannot be equal to the existing status ('%s' != '%s')",
+        newStatus, payment.getExternalPaymentStatus());
+
+    return payment.toBuilder()
+        .externalPaymentStatus(newStatus)
+        .authorisedTimestamp(getAuthorisedTimestamp(payment, newStatus))
+        .vehicleEntrantPayments(payment.getVehicleEntrantPayments()
```
plenza: Stream op to helper method

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
radekmorytko: Good point, added.

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
radekmorytko: Happy to help :)

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
radekmorytko: :+1:

src/main/java/uk/gov/caz/psr/service/GetAndUpdatePaymentsService.java
radekmorytko: It's not a silly question, you're right about it and not missing anything: from the user perspective it is not a valid case as this endpoint completes the payment journey, hence the payment has to be present in the database. The whole issue comes down to error propagation, whether we want to throw exception(s) or return some value. Initially I chose the latter, maybe because of the automatic inclinations that when a method starts with `get`, then `Optional` is a good candidate as a returned value. Currently controller checks if Optional is empty and returns `404` if it is, `200` code otherwise.

It's slightly a matter of taste, in my opinion. If you think the urgency of the error will be better reflected to the reader (I guess on the FE side they don't really care if we return `4xx` or `5xx` status code) by throwing an exception, I will be happy to change it.

src/main/java/uk/gov/caz/psr/service/PaymentExternalStatusFetcher.java
radekmorytko: You're right, I renamed the class and didn't change the javadoc, good catch, thanks :+1:

src/main/java/uk/gov/caz/psr/service/PaymentExternalStatusFetcher.java
radekmorytko: Fixed.

src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdater.java
radekmorytko: Sure.
Even though sometimes they can be useful, because they contain not only null checks, but also more complex predicates such as:
```java
Preconditions.checkArgument(!payment.getVehicleEntrantPayments().isEmpty(),
        "vehicle entrant payments cannot be empty");
Preconditions.checkArgument(status != payment.getExternalPaymentStatus(),
        "Status cannot be equal to the existing status ('%s' != '%s')",
        status, payment.getExternalPaymentStatus());
```
Anyway, I'll move those to a helper method.

src/main/java/uk/gov/caz/psr/service/GetAndUpdatePaymentsService.java
plenza: Thanks for very detailed explanation. It makes sense. I still can't get my head around how to properly handle exceptions in Java (I guess I will have to live my life with this uncertainty). Your solution is very good, and I'm OK with repositories being dump wrappers around DB - they don't see a row? They return Optional.empty. Simple. Caller is still obliged to handle it in a proper manner - maybe throw Exception, maybe skip etc. So ok for me!

src/it/resources/application-integration-tests.yml
```diff
@@ -5,7 +5,7 @@ spring:
     url: jdbc:postgresql://${postgres.url:localhost:9999}/caz-payments
 #    url: jdbc:postgresql://${postgres.url:localhost:5432}/caz-payments
     hikari:
-      maximum-pool-size: 10
+      maximum-pool-size: 1
```
plenza: This is working correctly in ITs?

plenza: 100% sure it timeouts in other projects when less than 5-6

radekmorytko: Yes it is

PR #43, opened 2019-11-14
[CAZ-1417] Repository layer for GET paymentStatus

musztardem:
**Introduced changes**
* Implemented `findBy(LocalDate, String)` method in `VehicleEntrantPaymentRepository`. Later this fetched record will be mapped to `PaymentStatusResponse`.
* Updated specs for implemented method.

**Additional info**

[Link to JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1417)


src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
```diff
@@ -53,6 +53,18 @@
       + "FROM vehicle_entrant_payment "
       + "WHERE travel_date = ? AND caz_id = ? AND vrn = ? AND payment_status = ?";
 
+  private static final String SELECT_PAYMENT_DETAILS_SQL = "SELECT "
```
plenza: SELECT_VEHICLE_ENTRANT_PAYMENT_BY_DATE_AND_VRN_SQL really

radekmorytko: 1. I'm afraid the provided query does not match the specified requirements: `GET v1/charge-settlement/payment-status` needs to return `paymentId` which is the _GOV.UK Pay unique payment identifier_. (I agree that the requirement in the parent task `CAZ-1327` is vague when it comes to this value: _paymentID is supposed to be derived from Payment table associated with selected Vehicle Entrant Payment entry_, but ICD is clear about this).
2. Taking into account the above and the database schema, my suggestions to the query are:
* initially we can try to find an entry in the `vehicle_entrant` table - it contains an index on `(caz_entry_date,caz_id,vrn)` (these are input params), so it is really well optimized
* secondly we can _inner join_ the `vehicle_entrant_payment`. This query will return a _non-null_ value if and only if the payment has been successfully made against `(caz_entry_date,caz_id,vrn)`
* lastly we can _inner join_ `payments` table to fetch `payment_provider_id`, which will be returned as `paymentId`.
The above query will return nothing if the payment was not successfully made.
3. The proposed query does not match any existing  repository, hence I think we can create a separate one, e.g. `PaymentStatusRepository`. The alternative is to split the query and first use `uk.gov.caz.psr.repository.VehicleEntrantRepository#findBy`, then find the matching `VehicleEntrantPayment` and, finally, `Payment`. The downside of this approach is the we will need to make two or three separate queries instead of one, but maybe the code will be cleaner? Waiting for your and others' opinion on this.
4. As Patryk [suggested](https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/43#pullrequestreview-317138286), it would be great to have integration tests for the repository. Later they can evolve to integration tests of the whole API endpoint, but for now it is a good starting point.

src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
```diff
@@ -53,6 +53,18 @@
...
+      + "payment_id, "
+      + "vehicle_entrant_payment_id, "
+      + "vrn, "
+      + "caz_id, "
+      + "travel_date, "
+      + "charge_paid, "
+      + "payment_status, "
+      + "case_reference "
+      + "FROM vehicle_entrant_payment "
+      + "WHERE travel_date = ? AND vrn = ?";
```
dratka: We should filter by CAZ_ID

src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
plenza: I would go with new repository for this query. `PaymentStatusRepository` sounds good to me.
And I would try to combine it in 1 SQL Select if not overly complicated (and I don't think it will be overly complicated).

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -0,0 +1,81 @@
...
+      + "FROM vehicle_entrant_payment "
+      + "INNER JOIN payment on vehicle_entrant_payment.payment_id = payment.payment_id "
+      + "WHERE vehicle_entrant_payment.caz_id = ? AND "
+      + "vehicle_entrant_payment.vrn = ? AND "
+      + "vehicle_entrant_payment.travel_date = ?";
+
+  /**
+   * Creates a new instance of this class.
+   */
+  public PaymentStatusRepository(JdbcTemplate jdbcTemplate) {
```
radekmorytko: You can use `@AllArgsConstructor` annotation here

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -0,0 +1,81 @@
...
+          preparedStatement.setObject(2, vrn);
+          preparedStatement.setObject(3, dateOfCazEntry);
+        },
+        ROW_MAPPER
+    );
+
+    return results;
+  }
+
+  private static class PaymentStatusRowMapper implements RowMapper<PaymentStatusResponse> {
```
radekmorytko: Please add javadoc

src/main/java/uk/gov/caz/psr/model/PaymentStatus.java
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * The unique payment ID coming from GOV UK Pay services.
+   */
+  String externalId;
+
+  /**
+   * Status of the payment.
+   */
+  ChargeSettlementPaymentStatus status;
```
radekmorytko: Instead of using a DTO class in model layer, please use `uk.gov.caz.psr.model.InternalPaymentStatus` - it is essentially the same

src/main/java/uk/gov/caz/psr/dto/ChargeSettlementPaymentStatus.java
```diff
@@ -1,8 +1,19 @@
...
   PAID,
   NOT_PAID,
   REFUNDED,
-  CHARGEBACK
+  CHARGEBACK;
+
+  /**
+   * Maps {@link ChargeSettlementPaymentStatus} to {@link InternalPaymentStatus}.
+   */
+  public static ChargeSettlementPaymentStatus from(InternalPaymentStatus internalPaymentStatus) {
```
radekmorytko: :+1:

src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
```diff
@@ -5,18 +5,34 @@
...
+import uk.gov.caz.psr.model.InternalPaymentStatus;
+import uk.gov.caz.psr.model.PaymentStatus;
 
 @Value
 @Builder
 public class PaymentStatusResponse {
   @ApiModelProperty(value = "${swagger.model.descriptions.payment-status.payment-status}")
   @NotNull
-  ChargeSettlementPaymentStatus chargeSettlementPaymentStatus;
+  InternalPaymentStatus paymentStatus;
```
radekmorytko: The previous version was good in this case - `PaymentStatusResponse` is a DTO and therefore should be composed of other DTOs :) Please restore the previous version and make use of `ChargeSettlementPaymentStatus#from` method in `PaymentStatusResponse#from` (line 31)

PR #44, opened 2019-11-15
[CAZ-1430] Add method findAllBy to VehicleEntrantPaymentRepository

dratka:
Added findAllBy to VehicleEntrantPaymentRepository to get all VehicleEntrantPayments for provided details.

### DIFF
--- a/src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
+++ b/src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
@@ -1,6 +1,8 @@
 package uk.gov.caz.psr.repository;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.time.LocalDate;
@@ -52,6 +54,36 @@ public class VehicleEntrantPaymentRepository {
       + "case_reference "
       + "FROM vehicle_entrant_payment "
       + "WHERE travel_date = ? AND caz_id = ? AND vrn = ? AND payment_status = ?";
+
+  @VisibleForTesting
+  static final String SELECT_BY_VRN_CAZ_ENTRY_DATE_AND_STATUS_SQL = "SELECT "
+      + "payment_id, "
+      + "vehicle_entrant_payment_id, "
+      + "vrn, "
+      + "caz_id, "
+      + "travel_date, "
+      + "charge_paid, "
+      + "payment_status, "
+      + "case_reference "
+      + "FROM vehicle_entrant_payment "
+      + "WHERE caz_id = ? AND vrn = ? AND travel_date = ? AND payment_status = ?";
+
+  @VisibleForTesting
+  static final String SELECT_BY_EXTERNAL_PAYMENT_VRN_AND_STATUS_SQL = "SELECT "
+      + "vehicle_entrant_payment.payment_id, "
+      + "vehicle_entrant_payment.vehicle_entrant_payment_id, "
+      + "vehicle_entrant_payment.vrn, "
+      + "vehicle_entrant_payment.caz_id, "
+      + "vehicle_entrant_payment.travel_date, "
+      + "vehicle_entrant_payment.charge_paid, "
+      + "vehicle_entrant_payment.payment_status, "
+      + "vehicle_entrant_payment.case_reference "
+      + "FROM vehicle_entrant_payment "
+      + "INNER JOIN payment ON vehicle_entrant_payment.payment_id = payment.payment_id "
+      + "WHERE vehicle_entrant_payment.caz_id = ? AND "
+      + "vehicle_entrant_payment.travel_date = ? AND "
+      + "vehicle_entrant_payment.payment_status = ? AND "
+      + "payment.payment_provider_id = ?";
 
   private static final String UPDATE_SQL = "UPDATE vehicle_entrant_payment "
       + "SET vehicle_entrant_id = ?, "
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
```diff
@@ -53,6 +53,33 @@
       + "FROM vehicle_entrant_payment "
       + "WHERE travel_date = ? AND caz_id = ? AND vrn = ? AND payment_status = ?";
 
+  private static final String FIND_ALL_BY_SQL = "SELECT "
```
plenza: BY_ what?

src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
```diff
@@ -208,6 +235,62 @@ public void update(VehicleEntrantPayment vehicleEntrantPayment) {
...
+   *
+   * @param cleanZoneId       provided Clean Air Zone ID
+   * @param cazEntryDate      Date of entry to provided CAZ
+   * @param externalPaymentId Payment Id from GOV.UK PAY
+   * @return list of found {@link VehicleEntrantPayment}.
+   * @throws NullPointerException     if {@code cleanZoneId} is null.
+   * @throws NullPointerException     if {@code cazEntryDate} is null.
+   * @throws IllegalArgumentException if {@code vrn} is empty.
+   */
+  public List<VehicleEntrantPayment> findAllBy(UUID cleanZoneId, LocalDate cazEntryDate,
```
plenza: Interesting case. I think that this operation should only return 0 or 1 `VehicleEntrantPayment` entities? Am I correct @radekmorytko ? 

This may be an opportunity to validate this global system invariant - if we have more than 1 we should throw an Exception and ultimately generate 500 Internal Server error. The sooner we report this error the better.

radekmorytko: > Interesting case. I think that this operation should only return 0 or 1 VehicleEntrantPayment entities?

Not necessarily, because one payment (`external_payment_id`) can be for more than one day (= more than one `VehicleEntrantPayment` entity). Currently this query can return up to `13` rows if I'm counting correctly (one payment can be done for today plus/minus six days).

dratka: It is an interesting idea to throw an exception here. I was planning to use this method to handle exceptions in service to avoid throwing exceptions for business logic in the repository. 

@radekmorytko you are right but I added cazEntryDate as a param because we need to find VehicleEntratnPayment for the chosen day.

radekmorytko: When I'm thinking of the whole scenario now, I think that we do not validate anywhere the status of the payment. As it is stated in the ICD:
> This PUT method allows Local Authorities to update the status of one or more **paid** CAZ

**paid** implies that the payment exists and is in `SUCCESS` state.
Maybe we can add `findSuccessfuPaymentByExternalId` in `PaymentRepository` instead of having this method?

dratka: So we should only update the status for paid vehicleEntrantPayments? I thought that they will be able to update it back to paid if they made a mistake. But you can be right that ICD sais that it is only for paid CAZ.

plenza: Good question. ICD is ambiguous. Firstly it states:

image

But then:

image

In my opinion we should allow to update to `Paid` or `NotPaid`. This is because motorist can pay directly in LA office, or by SMS etc. It won't go through our service or Gov.uk Pay.

But to be sure we should ask @szromek and discuss it a bit.

src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
```diff
@@ -208,11 +239,87 @@ public void update(VehicleEntrantPayment vehicleEntrantPayment) {
...
+   * @throws NullPointerException     if {@code cleanZoneId} is null.
+   * @throws NullPointerException     if {@code cazEntryDate} is null.
+   * @throws IllegalArgumentException if {@code vrn} is empty.
+   * @throws RuntimeException         if method found more than one VehicleEntrantPayment.
+   */
+  public Optional<VehicleEntrantPayment> findOnePaidByCazEntryDateAndExternalPaymentId(
+      UUID cleanZoneId, LocalDate cazEntryDate, String externalPaymentId) {
+    Preconditions.checkNotNull(cleanZoneId, "cleanZoneId cannot be null");
+    Preconditions.checkNotNull(cazEntryDate, "cazEntryDate cannot be null");
+    Preconditions.checkArgument(!externalPaymentId.isEmpty(), "externalPaymentId cannot be empty");
```
radekmorytko: I think we need to make sure `externalPaymentId` is not null first (or we can use guava's [isNullOrEmpty](https://guava.dev/releases/20.0/api/docs/com/google/common/base/Strings.html#isNullOrEmpty-java.lang.String-)). The same applies for the other method.

src/main/java/uk/gov/caz/psr/repository/VehicleEntrantPaymentRepository.java
```diff
@@ -208,11 +239,87 @@ public void update(VehicleEntrantPayment vehicleEntrantPayment) {
...
+   * {@link Optional#empty()} is returned. If more then one found throws {@code RuntimeException}.
+   *
+   * @param cleanZoneId       provided Clean Air Zone ID
+   * @param cazEntryDate      Date of entry to provided CAZ
+   * @param externalPaymentId Payment Id from GOV.UK PAY
+   * @return list of found {@link VehicleEntrantPayment}.
+   * @throws NullPointerException     if {@code cleanZoneId} is null.
+   * @throws NullPointerException     if {@code cazEntryDate} is null.
+   * @throws IllegalArgumentException if {@code vrn} is empty.
+   * @throws RuntimeException         if method found more than one VehicleEntrantPayment.
```
radekmorytko: Wrong exception's type, should be `IllegalStateException`

PR #50, opened 2019-11-19
Add utility class to parse pennies into pounds

kingarj:


src/main/java/uk/gov/caz/psr/util/CurrencyFormatter.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.caz.psr.util;
```
radekmorytko: Minor - please add javadoc both for the class and the method (client's requirement)

kingarj: @radekmorytko this has now been fixed.

PR #56, opened 2019-11-25
[CAZ-1494] Return emailAddress from GOV.UK PAY when checking payment status.

dratka:
Jira card: https://eaflood.atlassian.net/browse/CAZ-1494

Updated the endpoint to return emailAddress in getExternalPaymentAndUpdateStatus endpoint. 

Decided with @radekmorytko to remove OnBeforePublishPaymentStatusUpdateEvent callback to make the code simpler.
Added a new model to collect ExternalPaymentData inside services. 
Updated all Unit and IT tests.


src/main/java/uk/gov/caz/psr/dto/GetAndUpdatePaymentStatusResponse.java
```diff
@@ -13,14 +14,19 @@
 public class GetAndUpdatePaymentStatusResponse {
   @NonNull
   UUID paymentId;
+
   @NonNull
   SuccessFailurePaymentStatus status;
 
+  @NotNull
```
radekmorytko: Unfortunately this field is `nullable` - for payments which have not finished successfully it can contain `null`.

src/main/java/uk/gov/caz/psr/model/ExternalPaymentDetails.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * An entity which contains details of Payment in GOV.UK PAY.
+ */
+@Value
+@Builder
+public class ExternalPaymentDetails {
+  /**
+   * Status of payment in GOV.UK PAY
+   */
+  ExternalPaymentStatus externalPaymentStatus;
```
radekmorytko: We can annotate it here with `lombok.NonNull` annotation.

src/main/java/uk/gov/caz/psr/model/ExternalPaymentDetails.java
```diff
@@ -0,0 +1,21 @@
...
+
+  /**
+   * Email registered in GOV.UK PAY during the payment.
```
radekmorytko: I would explicitly add a comment stating this field is nullable.

src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdater.java
```diff
@@ -16,27 +16,26 @@
...
   private final TransientVehicleEntrantsLinker transientVehicleEntrantsLinker;
   private final ApplicationEventPublisher applicationEventPublisher;
 
   /**
    * Updates {@code payment} with a new {@code status}.
    */
-  public Payment updateWithStatus(Payment payment, ExternalPaymentStatus status,
-      OnBeforePublishPaymentStatusUpdateEvent onBeforePublishEvent) {
-    checkPreconditions(payment, status, onBeforePublishEvent);
+  public Payment updateWithStatus(Payment payment, ExternalPaymentDetails externalPaymentDetails) {
```
radekmorytko: What do you think about renaming this method to `updateWithExternalDetails`?

src/main/java/uk/gov/caz/psr/dto/external/GetPaymentResult.java
```diff
@@ -47,5 +49,17 @@ public ExternalPaymentStatus getPaymentStatus() {
       return ExternalPaymentStatus.UNKNOWN;
     }
   }
+
+  /**
+   * Maps this value object to an instance of the model.
+   *
+   * @return An instance of {@link ExternalPaymentDetails} whose parameters comes from this object.
+   */
+  public ExternalPaymentDetails toExternalPaymentDetails() {
```
plenza: It's a bit bad idea to pollute DTOs with knowledge about model classes. It happens that usually we move some external service logic, like in this case Gov.uk pay service, including call and DTOs to some other place/library. So DTOs should be as dumb as possible. 

What I would do? I'd move conversion method to model class and call it `static from(GetPaymentResult..` or even better create some external converter class. Can be static utility, we won't need to mock or inject it.

src/main/java/uk/gov/caz/psr/service/GetAndUpdatePaymentsService.java
```diff
@@ -50,17 +50,19 @@
...
     GetPaymentResult paymentInfo = externalPaymentsRepository.findById(externalPaymentId)
         .orElseThrow(() -> new IllegalStateException("External payment not found with id "
-                + "'" + externalPaymentId + "'"));
+            + "'" + externalPaymentId + "'"));
+    ExternalPaymentDetails externalPaymentDetails = paymentInfo.toExternalPaymentDetails();
 
-    ExternalPaymentStatus externalStatus = paymentInfo.getPaymentStatus();
-
-    if (payment.getExternalPaymentStatus() == externalStatus) {
+    if (payment.getExternalPaymentStatus() == externalPaymentDetails.getExternalPaymentStatus()) {
```
plenza: `if (sameStatuses(payment, externalPaymentDetails)`

PR #59, opened 2019-11-25
[CAZ-1429] Fixes bug with not updating case_reference in PUT payment_status update action

dratka:
QA found a bug that they are not able to update case_reference value during payment_status update.
I've found and fixed the bug that we are not assigning it to the update objects.


src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdateService.java
```diff
@@ -71,6 +71,7 @@ private VehicleEntrantPayment prepareVehicleEntrantPayment(
 
     return vehicleEntrantPayment.toBuilder()
         .internalPaymentStatus(vehicleEntrantPaymentStatusUpdate.getPaymentStatus())
+        .caseReference(vehicleEntrantPaymentStatusUpdate.getCaseReference())
```
radekmorytko: Can we add a test case for this bug?

PR #60, opened 2019-11-26
add db migration endpoint

phan-anh-tuan:


src/main/java/uk/gov/caz/psr/controller/DbSchemaMigrationControllerApi.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import uk.gov.caz.psr.util.LiquibaseWrapper;
+
+@RestController
+public class DbSchemaMigrationControllerApi implements DbSchemaMigrationControllerApiSpec {
+  public static final String PATH = "/v1/migrate-db";
+
+  @Autowired
+  LiquibaseWrapper liquibaseWrapper;
```
cruddasj: @phan-anh-tuan - can we please avoid using the autowired annotation? The remainder of the codebase uses constructor injection which Spring supports.

For the tests - it would also make sense to mirror those seen in the VCCS service...

phan-anh-tuan: @cruddasj I addressed your comment on `@Autowired` annotation but please can you clarify the comment regarding the controller test?

cruddasj: Hey. If you look at https://github.com/InformedSolutions/JAQU-CAZ-Vehicle-Compliance-Checker-API/commit/a0846fd28c66fc7bf745dd4e16bc26516c05e55d you'll be able to see the integration test for the liquibase wrapper. Can we please make sure that the test is present for all services and constructor injection is used across the board (e.g. for tariffs, NTR, payments, retrofit etc.)?

pom.xml
```diff
@@ -32,6 +32,7 @@
       <!-- TODO remove once ChargeSettlementController is implemented and tested -->
       **/psr/dto/ChargeSettlementPaymentStatus.java,
       **/psr/controller/ChargeSettlementController.java,
+      **/util/LiquibaseWrapper.java,
```
plenza: Both excludes are most probably not needed because there are tests?

phan-anh-tuan: The integration test did not verify the rollback scenario hence probably (removing that exclusion) will fail the test coverage check.

phan-anh-tuan: Exclusions are removed and tests are solidified.

PR #74, opened 2019-12-02
[CAZ-1349] Handle Error path for PUT payment_status update endpoint

dratka:
Added support for Error path in PUT charge_settelment#paument_status action. 
Added support for Request validation errors in response. 
Handled all possible errors
Added tests.

### DIFF
--- a/src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
+++ b/src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
@@ -4,7 +4,9 @@ import java.util.List;
 import java.util.UUID;
 import lombok.AllArgsConstructor;
 import org.springframework.http.ResponseEntity;
+import org.springframework.validation.BindingResult;
 import org.springframework.web.bind.annotation.RestController;
+import uk.gov.caz.psr.controller.exception.DtoValidationException;
 import uk.gov.caz.psr.dto.PaymentInfoRequest;
 import uk.gov.caz.psr.dto.PaymentInfoResponse;
 import uk.gov.caz.psr.dto.PaymentStatusRequest;
@@ -60,8 +62,11 @@ public class ChargeSettlementController implements ChargeSettlementControllerApi
 
   @Override
   public PaymentUpdateSuccessResponse updatePaymentStatus(PaymentStatusUpdateRequest request,
-      String apiKey) {
+      BindingResult bindingResult, String apiKey) {
     UUID cleanAirZoneId = UUID.fromString(apiKey);
+    if (bindingResult.hasErrors()) {
+      throw new DtoValidationException(request.getVrn(), bindingResult);
+    }
     paymentStatusUpdateService
         .processUpdate(request.toVehicleEntrantPaymentStatusUpdates(cleanAirZoneId));
     return new PaymentUpdateSuccessResponse();
--- a/src/main/java/uk/gov/caz/psr/controller/ExceptionController.java
+++ b/src/main/java/uk/gov/caz/psr/controller/ExceptionController.java
@@ -1,0 +1,75 @@
+package uk.gov.caz.psr.controller;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.ResponseEntity;
+import org.springframework.validation.FieldError;
+import org.springframework.web.bind.annotation.ExceptionHandler;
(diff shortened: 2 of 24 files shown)


src/main/java/uk/gov/caz/psr/controller/exception/DtoValidationException.java
```diff
@@ -0,0 +1,26 @@
+package uk.gov.caz.psr.controller.exception;
+
+import org.springframework.validation.BindingResult;
+
+/**
+ * Exception class which will be used to throw exception when request validation fails in charge
+ * settlement API.
+ */
+public class DtoValidationException extends IllegalArgumentException {
```
radekmorytko: The class can be simplified to:
```java
@Value
public class DtoValidationException extends IllegalArgumentException {
  String vrn;
  BindingResult bindingResult;
}
```

src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
```diff
@@ -24,6 +27,7 @@
  */
 @RestController
 @AllArgsConstructor
+@Slf4j
```
radekmorytko: I think this is a leftover as there is no log statement in the changes and can be safely removed.

src/main/java/uk/gov/caz/psr/dto/ErrorResponse.java
```diff
@@ -0,0 +1,52 @@
...
+@Builder
+public class ErrorResponse {
+
+  private static final String NO_VRN = "";
+  private static final String VALIDATION_ERROR_TITLE = "Validation error";
+
+  String vrn;
+  String title;
+  String detail;
+  Integer status;
```
radekmorytko: I think we can change `Integer` to `int` as we don't expect `null` values here.

src/main/java/uk/gov/caz/psr/dto/ErrorResponse.java
```diff
@@ -0,0 +1,52 @@
...
+  public static ErrorResponse validationErrorResponseWithDetailAndVrn(String vrn, String detail) {
+    return ErrorResponse.builder()
+        .vrn(vrn)
+        .title(VALIDATION_ERROR_TITLE)
+        .detail(detail)
+        .status(HttpStatus.BAD_REQUEST.value())
+        .build();
+  }
+
+  public int getStatus() {
```
radekmorytko: This _getter_ can be removed as we have `@Value` on the class.

src/main/java/uk/gov/caz/psr/dto/ErrorsResponse.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Collections;
+import java.util.List;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+import org.springframework.http.HttpStatus;
+
+@Value
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class ErrorsResponse {
```
radekmorytko: Missing javadoc

src/main/java/uk/gov/caz/psr/repository/exception/NotUniqueVehicleEntrantPaymentFoundException.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.caz.psr.repository.exception;
+
+/**
+ * Exception class which will be used to throw exception when {@code
+ * VehicleEntrantPaymentRepository} found more then one VehicleEntrantPayment for FindOne methods.
+ */
+public class NotUniqueVehicleEntrantPaymentFoundException extends RuntimeException {
```
radekmorytko: Maybe use `@Value`?

src/main/java/uk/gov/caz/psr/controller/ExceptionController.java
```diff
@@ -0,0 +1,72 @@
...
+import uk.gov.caz.psr.dto.ErrorResponse;
+import uk.gov.caz.psr.dto.ErrorsResponse;
+import uk.gov.caz.psr.model.ValidationError;
+import uk.gov.caz.psr.repository.exception.NotUniqueVehicleEntrantPaymentFoundException;
+import uk.gov.caz.psr.service.exception.MissingVehicleEntrantPaymentException;
+
+@Slf4j
+@RestControllerAdvice
+@Order(Ordered.HIGHEST_PRECEDENCE)
+public class ExceptionController extends GlobalExceptionHandler {
```
radekmorytko: Can we add logging in every method that an exception occurred?

PR #79, opened 2019-12-03
Feature - CAZ:1318 - Multiple Gov.UK Pay Accounts

kingarj:
- Adds a class for interacting with AWS Secrets Manager
- Adds logic to redirect payments to separate Gov.UK Pay accounts (using API keys stored in Secrets Manager)


src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
...
+   * 
+   * @return secret the contents of the secret
+   */
+  public String getSecretsValue() {
+    GetSecretValueRequest getSecretValueRequest =
+        new GetSecretValueRequest().withSecretId(generateSecretName());
+    GetSecretValueResult getSecretValueResult = null;
+    getSecretValueResult = client.getSecretValue(getSecretValueRequest);
+
+    // Decrypts secret using the associated KMS CMK.
```
plenza: I don't think any decryption happens here. Comment seems to be misleading

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
```diff
@@ -36,12 +42,32 @@ public void processDanglingPayment(Payment danglingPayment) {
 
     Stopwatch stopwatch = Stopwatch.createStarted();
     try {
+      danglingPayment = loadVehicleEntrantPayments(danglingPayment);
+
+      Optional<UUID> cleanAirZoneId =
```
radekmorytko: The similar logic is duplicated two times (apart from here):
* `uk.gov.caz.psr.service.InitiatePaymentService#createPayment`:
```java
// Retrieve API key for appropriate Gov.UK Pay account
Optional<String> apiKey = credentialRetrievalManager.getApiKey(request.getCleanAirZoneId());
if (apiKey.isPresent()) {
  externalPaymentsRepository.setApiKey(apiKey.get());
} else {
  throw new NoSuchElementException(
      "The API key has not been set for the given Clean Air Zone ID.");
}
```
* `uk.gov.caz.psr.service.GetAndUpdatePaymentsService#getExternalPaymentAndUpdateStatus`

My suggestion:
The fact that something like API key exists is an implementation detail of `ExternalPaymentsRepository`. Additionally, we will get rid of the duplication when we move the logic of retrieving the API key for the CAZ (as processing a payment in PSR is always in the context of a *single* CAZ) to `ExternalPaymentsRepository`. What are the necessary changes? There are two methods that form its API:
* `uk.gov.caz.psr.repository.ExternalPaymentsRepository#create` that accepts `Payment`. This object must contain non-empty `vehicleEntrantPayments` (we can add a precondition to this method to be sure) and we can get _cazId_ from those
* `uk.gov.caz.psr.repository.ExternalPaymentsRepository#findById` that accepts an internal payment identifier. We can add a second argument, `UUID cleanAirZoneId`, and rename the method to something like `findByIdFor` or `findByIdAndCazId`

Additionally, when the API key is not found we should fail miserably as it means the service is misconfigured and cannot serve its purpose.

Example:
```java
// for create
RequestEntity<CreateCardPaymentRequest> request =
    buildRequestEntityForCreate(getApiKeyFor(payment), buildCreateBody(payment, returnUrl))
...

private RequestEntity<CreateCardPaymentRequest> buildRequestEntityForCreate(
    String apiKey, CreateCardPaymentRequest body) {
  return RequestEntity.post(createPaymentUri)
      .header("Authorization", "Bearer " + apiKey)
      .accept(MediaType.APPLICATION_JSON)
      .contentType(MediaType.APPLICATION_JSON)
      .body(body);
}
...

// for get
RequestEntity<Void> request = buildRequestEntityForFindById(getApiKeyFor(cleanZoneId), id);
...
private RequestEntity<Void> buildRequestEntityForFindById(String apiKey, String id) {
  return RequestEntity.get(buildFindByUri(id))
      .header("Authorization", "Bearer " + apiKey)
      .accept(MediaType.APPLICATION_JSON)
      .build();
}
...

// helper methods

private String getApiKeyFor(Payment payment) {
  // invariant 1: payment.getVehicleEntrantPayments() is not empty (can be enforced by Preconditions.checkArgument(...) at the beginning of _create_ method)
  // invariant 2: payment.getVehicleEntrantPayments() constains the same caz id
  UUID cleanZoneId = payment.getVehicleEntrantPayments().iterator().next()
      .getCleanZoneId();
  return getApiKeyFor(cleanZoneId);
}

private String getApiKeyFor(UUID cleanZoneId) {
  String apiKey = credentialRetrievalManager.getApiKey(cleanZoneId)
      .orElseThrow(() -> new IllegalStateException("The API key has not been set for Clean Air "
          + "Zone " + cleanZoneId));
  logMaskedApiKey(apiKey, cleanZoneId);
  return apiKey;
}

// now it accepts _cleanZoneId_ as _apiKey_ is not unique
private void logMaskedApiKey(String apiKey, UUID cleanZoneId) {
  int toReveal = Math.min(3, apiKey.length());
  int toMask = apiKey.length() - toReveal;
  log.info("GOV UK PAY api key for caz '{}': {}{}", cleanZoneId, apiKey.substring(0, toReveal),
      Strings.repeat("*", toMask));
}

```

What are some of the implications of the proposed changes?
* the logic in `uk.gov.caz.psr.service.InitiatePaymentService#createPayment`, `uk.gov.caz.psr.service.GetAndUpdatePaymentsService#getExternalPaymentAndUpdateStatus` and `uk.gov.caz.psr.service.CleanupDanglingPaymentService#processDanglingPayment` stays nearly the same as it was and we do not have the duplications
* we can remove `uk.gov.caz.psr.repository.ExternalPaymentsRepository#setApiKey`. In general, _services_ in nearly _every case_ should be _stateless_ (e.g. so that they can be easily reused in a multithreaded environment). `ExternalPaymentsRepository` should be stateless as well

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
...
+   * Constructor for manager class of the external secrets repository.
+   * 
+   * @param secretPrefix the prefix of the secret
+   * @param secretName the name of the secret
+   * @param secretProfileSeparator the profile separator of the secret
+   * @param secretEnvironment the environment of the secret
+   * @param awsSecretsManager an instance of {@link AWSSecretsManager}
+   * @param objectMapper an instance of {@link ObjectMapper}
+   */
+  public CredentialRetrievalManager(@Value("${aws.secretsmanager.prefix}") String secretPrefix,
```
radekmorytko: Those configuration properties are related to each other and can be grouped into a _springified_ and _lombokised_ POJO, e.g.
```java
@Component
@ConfigurationProperties(prefix = "aws.secretsmanager")
@Data
public class SecretsManagerProperties {
  String prefix;
  String name;
  String profileSeparator;
  String environment;
}
```
We can inject it then in the constructor as a regular Spring bean.

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
...
+    String cleanAirZoneStr = cleanAirZoneId.toString();
+    return cleanAirZoneStr.replace("-", "");
+  }
+
+  /**
+   * Helper method to generate a full secret name from a number of variables.
+   * 
+   * @return the full secret name
+   */
+  private String generateSecretName() {
```
radekmorytko: Maybe instead of generating it multiple times we can do it once in the constructor, assign the variable to a `private final` instance field and name it `secretId`?

pom.xml
```diff
@@ -256,6 +256,18 @@
...
         <executions>
+          <execution>
+            <id>add-source</id>
+            <phase>generate-sources</phase>
+            <goals>
+                <goal>add-source</goal>
+            </goals>
+            <configuration>
+                <sources>
+                    <source>${project.build.directory}/generated-sources/annotations/</source>
```
plenza: @kingarj Why did you add this `add-source` goal? Something is not building properly? No one has problems with that and drone is indeed successfully building the project without it. ?

pom.xml
```diff
@@ -404,6 +416,7 @@
           <failOnViolation>true</failOnViolation>
           <linkXRef>false</linkXRef>
           <violationSeverity>warning</violationSeverity>
+          <excludes>**/psr/model/info/*_.java</excludes>
```
plenza: Files are generated to ./target/generated-sources directory, as by Maven specification. As so they are not "taken" by Sonar or Jacoco for analysis.

.gitignore
```diff
@@ -31,9 +31,12 @@ HELP.md
 ### Vagrant ###
 .vagrant
 
+### Hibernate ###
+*_.java
```
plenza: Files are generated to ./target/generated-sources as by Maven specification. ./target is already ignored.

src/it/java/uk/gov/caz/psr/DanglingPaymentsCleanupTestIT.java
```diff
@@ -88,28 +123,19 @@ public void deleteQueue() {
...
-    thenStatusOf("cancelled-payment-id")
-        .isFailed()
-        .andStatusOfVehicleEntrantPaymentsIsNotPaid();
-    thenStatusOf("expired-payment-id")
-        .isFailed()
-        .andStatusOfVehicleEntrantPaymentsIsNotPaid();
-    thenStatusOf("success-payment-id")
-        .isSuccess()
-        .andStatusOfVehicleEntrantPaymentsIsPaid()
+    thenStatusOf("cancelled-payment-id").isFailed().andStatusOfVehicleEntrantPaymentsIsNotPaid();
```
plenza: Misconfigured formatter is causing changes in already well formatted code and making this code review quite a bit harder to analyse. Could you please configure formatter to match our style? There is Eclipse formatter style file in ./developer-resources sub-directory.

src/it/java/uk/gov/caz/psr/DanglingPaymentsCleanupTestIT.java
```diff
@@ -75,6 +88,28 @@ public void startMockServer() {
     mockServer = startClientAndServer(1080);
   }
 
+  @BeforeEach
+  public void createSecret() throws JsonProcessingException {
+    String cazIdFormatted = "53e03a28-0627-11ea-9511-ffaaee87e375".replace("-", "");
```
plenza: DanglingPaymentsCleanupTest is not related to stuff connected to SecretsManager yet currently it is polluted by code required to setup SM. It will be needed in (most probably) all other integration tests. Could you please extract code needed to initialize and prepare SM into some util - helper class and only call it here (1 line)

src/it/java/uk/gov/caz/psr/GetAndUpdatePaymentStatusTestIT.java
```diff
@@ -53,9 +52,10 @@ public void shouldReturn404StatusWhenPaymentDoesNotExistInDatabase() throws Exce
     UUID notExistingId = UUID.fromString("8916a4e0-fb9e-11e9-8483-b3fce09a0a00");
 
     String correlationId = "542de1b5-4aab-45eb-bccc-6ec91f1d6d51";
-    mockMvc.perform(get(URL_TEMPLATE, notExistingId)
-        .header(Constants.X_CORRELATION_ID_HEADER, correlationId)
-        .accept(MediaType.APPLICATION_JSON))
+    mockMvc
```
plenza: Formatter issue messed up already correctly formatted lines.

src/it/java/uk/gov/caz/psr/SuccessPaymentsJourneyTestIT.java
```diff
@@ -96,6 +107,28 @@ public void createEmailQueue() {
     sqsClient.createQueue(createQueueRequest);
   }
 
+  @BeforeEach
+  public void createSecret() throws JsonProcessingException {
```
plenza: Same as [here](https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/79/files#r354300053)

src/main/java/uk/gov/caz/psr/Application.java
```diff
@@ -5,13 +5,14 @@
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 
 @SpringBootApplication(scanBasePackages = {
-    "uk.gov.caz.psr.configuration",
+    "uk.gov.caz.psr.configuration",
```
plenza: Redundant space

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
plenza: Then constructor can be removed and class annotated by `@AllArgsConstructor`

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
...
+    this.objectMapper = objectMapper;
+
+  }
+
+  /**
+   * Retrieve the appropriate API key given a Clean Air Zone identifier.
+   * 
+   * @return secret the contents of the secret
+   */
+  public String getSecretsValue() {
```
plenza: Please rename method to something better reflecting our intent: 'retrieveCazGovUkPayApiKey` maybe?

plenza: There is not need for this method to be public - it is only internal helper of `getApiKey`. Regarding the tests - it must be tested implicitly by tests of `getApiKey`

radekmorytko: As far as I understand this method returns _all_ secrets held in _secretsmanager_. What is the expected format of storage? A JSON object with simple mapping `cazId => api key`? If this is the case, I'd recommend to change this method to return `Map<String, String>` instead of `String`:
```java
String secretString = getSecretValueResult.getSecretString() != null
    ? getSecretValueResult.getSecretString()
    : new String(Base64.getDecoder().decode(getSecretValueResult.getSecretBinary()).array());

try {
  return objectMapper.readValue(secretString, new TypeReference<Map<String, String>>() {});
} catch (JsonProcessingException e) {
  log.error("Error while parsing AWS secrets:", e);
  return Collections.emptyMap();
}
```

The the code in `getApiKey` is simpler:
```java
public Optional<String> getApiKey(UUID cleanAirZoneId) {
  Map<String, String> secretsValue = getSecretsValue();
  String secretName = generateSecretKey(cleanAirZoneId);
  if (secretsValue.containsKey(secretName)) {
    log.info("Successfully retrieved API key for Clean Air Zone: {}", cleanAirZoneId);
    return Optional.of(secretsValue.get(secretName));
  }
  log.error("Failed to retrieved API key for Clean Air Zone: {}", cleanAirZoneId);
  return Optional.empty();
}
```

src/main/java/uk/gov/caz/psr/model/service/VehicleEntrantPaymentsService.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.stream.Collectors;
+import org.springframework.stereotype.Service;
+import uk.gov.caz.psr.model.VehicleEntrantPayment;
+
+/**
+ * Service class for VehicleEntrantPayment objects.
+ * 
+ */
+@Service
+public class VehicleEntrantPaymentsService {
```
radekmorytko: If you follow my advice from other comments I think there's a chance this class may not be needed anymore

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  /**
+   * Get secret value by a given key from external property source.
+   * 
+   * @param cleanAirZoneId the ID of the Clean Air Zone which gives the key of the secret
+   * @return Secret value for a given key.
+   */
+  public Optional<String> getApiKey(UUID cleanAirZoneId) {
```
plenza: I don't think that returning Optional here is desirable. What shall happen if the key is missing or invalid? I think we should throw exception and report it as soon as possible. This is clearly serious error on our/admin side and it should not be possible for Payments service to function without these secrets.

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
+package uk.gov.caz.psr.domain.authentication;
```
radekmorytko: This package seems a bit weird. If we want to follow DDD convention, I think this still does not belong to `domain` - it retrieves data from an _external_ source, so this is still an application service, not the one from domain. I'd move it to `uk.gov.caz.psr.service.authentication` package.

src/main/java/uk/gov/caz/psr/domain/authentication/CredentialRetrievalManager.java
```diff
@@ -0,0 +1,115 @@
...
+      jsonNode = objectMapper.readTree(rawSecretString);
+      if (jsonNode.has(secretName)) {
+        log.info("Successfully retrieved API key for Clean Air Zone: {}", secretName);
+        String apiKey = jsonNode.get(secretName).textValue();
+        return Optional.ofNullable(apiKey);
+      } else {
+        return Optional.empty();
+      }
+    } catch (IOException e) {
+      return Optional.empty();
```
plenza: We must have `log.error` here - these kind of parsing errors will happen and without logs we are blind

pom.xml
kingarj: @plenza Eclipse was struggling to compile the *_ files and so to debug in the IDE I had to add this in.

kingarj:
Thanks for all your feedback @plenza @radekmorytko, I'll be looking to address this the rest of today and tomorrow.

src/it/java/uk/gov/caz/psr/DanglingPaymentsCleanupTestIT.java
kingarj: @plenza that's the one I'm using - unfortunately I think there might be some discrepencies between the two formatters. Might be worth having a look at it otherwise we may continue running into similar issues.

src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
```diff
@@ -38,29 +38,41 @@
...
+      CredentialRetrievalManager credentialRetrievalManager) {
     this.rootUrl = rootUrl;
-    this.restTemplate = restTemplateBuilder.interceptors(apiKeyHeaderInjector(apiKey)).build();
-    this.createPaymentUri = URI.create(rootUrl + CREATE_URI);
-    logMaskedApiKey(apiKey);
+    this.restTemplate = restTemplateBuilder.build();
+    this.credentialRetrievalManager = credentialRetrievalManager;
+  }
+
+  private String getApiKeyFor(Payment payment) {
```
radekmorytko: [minor] Please add a tiny bit of javadoc here and for the other `getApiKeyFor`

src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
```diff
@@ -69,27 +81,28 @@ private void logMaskedApiKey(String apiKey) {
...
+          restTemplate.exchange(request, CreatePaymentResult.class);
       CreatePaymentResult responseBody = responseEntity.getBody();
-      ExternalPaymentStatus externalPaymentStatus = toModelStatus(
-          responseBody.getState().getStatus());
-      return payment.toBuilder()
-          .externalId(responseBody.getPaymentId())
+      ExternalPaymentStatus externalPaymentStatus =
+          toModelStatus(responseBody.getState().getStatus());
+      return payment.toBuilder().externalId(responseBody.getPaymentId())
+          .submittedTimestamp(LocalDateTime.now()).externalPaymentStatus(externalPaymentStatus)
```
radekmorytko: There is a minor leftover from the merge with the develop - setting `submittedTimestamp` is done twice. The one can be safely removed.

src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
```diff
@@ -121,17 +133,17 @@ private ExternalPaymentStatus toModelStatus(String status) {
    * Gets payment details by its identifier.
    *
    * @param id ID of the payment.
```
radekmorytko: Can you please add javdoc for `cleanAirZoneId` param?

src/main/java/uk/gov/caz/psr/service/CleanupDanglingPaymentService.java
```diff
@@ -36,12 +39,20 @@ public void processDanglingPayment(Payment danglingPayment) {
 
     Stopwatch stopwatch = Stopwatch.createStarted();
     try {
+      danglingPayment = loadVehicleEntrantPayments(danglingPayment);
+      UUID paymentId = danglingPayment.getId();
+
+      UUID cleanAirZoneId =
```
radekmorytko: Can you please move it to a private, helper method and possibly even inline it in line 52 if it looks ok? Don't forget about javadoc! :)

src/main/java/uk/gov/caz/psr/model/service/VehicleEntrantPaymentsService.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Uses a list of vehicle entrant payments (associated with a payment) to find the Clean Air Zone
+   * ID of the payment.
+   * 
+   * @param vehicleEntrantPayments a list of VehicleEntrantPayment objects
+   * @return a unique identifier for a Clean Air Zone
+   */
+  public Optional<UUID> findCazId(List<VehicleEntrantPayment> vehicleEntrantPayments) {
```
radekmorytko: Currently when saving the payment in the database it is not possible to have `VehicleEntrantPayment` entities with different caz ids - therefore I'm wondering whether we need this method or just we can use
```
vehicleEntrantPayments.iterator().next().getCleanZoneId()
```
instead :thinking: 

If you're sure we want to have this check, my advice is to slightly simplify the code:
```java
public Optional<UUID> findCazId(List<VehicleEntrantPayment> vehicleEntrantPayments) {
  Set<UUID> cazIds = vehicleEntrantPayments.stream()
      .map(VehicleEntrantPayment::getCleanZoneId)
      .collect(Collectors.toSet());
  return cazIds.size() == 1 ? Optional.of(cazIds.iterator().next()) : Optional.empty();
}
```

src/test/java/uk/gov/caz/psr/service/CleanupDanglingPaymentServiceTest.java
```diff
@@ -139,12 +131,11 @@ void externalPaymentRepositoryNotCalledWhenCleanAirZoneIdNotFound() {
         ExternalPaymentDetailsFactory.anyWithStatus(ExternalPaymentStatus.INITIATED);
     Payment payment =
         paymentWithEmptyVehicleEntrantsAndStatus(externalPaymentDetails.getExternalPaymentStatus());
-    mockVehicleEntrantsFor(payment, UUID.randomUUID());
-    mockCazId(null);
+//    mockVehicleEntrantsFor(payment, UUID.randomUUID());
```
radekmorytko: Can safely be removed :)

PR #85, opened 2019-12-08
[CAZ-1415] Error scenarios for GET payments-status

musztardem:
**Introduced changes**
* Implemented error handling for `GET PaymentStatus`
* Adjusted tests

**Additional info**

[Link to JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1415)

**Testing**
* Tested all possible scenarios with Postman

Screenshot 2019-12-08 at 20 23 46

### DIFF
--- a/src/main/java/uk/gov/caz/psr/dto/PaymentStatusRequest.java
+++ b/src/main/java/uk/gov/caz/psr/dto/PaymentStatusRequest.java
@@ -1,12 +1,10 @@
 package uk.gov.caz.psr.dto;
 
 import io.swagger.annotations.ApiModelProperty;
-import java.time.LocalDate;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 import lombok.Value;
-import org.springframework.format.annotation.DateTimeFormat;
-import org.springframework.format.annotation.DateTimeFormat.ISO;
 
 @Value
 public class PaymentStatusRequest {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/caz/psr/dto/PaymentStatusRequest.java
```diff
@@ -18,6 +16,6 @@
 
   @ApiModelProperty(value = "${swagger.model.descriptions.payment-status.date-of-caz-entry}")
   @NotNull
-  @DateTimeFormat(iso = ISO.DATE)
-  LocalDate dateOfCazEntry;
+  @Pattern(regexp =  "^\\d{4}-\\d{2}-\\d{2}$")
```
plenza: `ISO.DATE` is the same as this regex? Why did you change it?

src/main/java/uk/gov/caz/psr/dto/PaymentStatusRequest.java
```diff
@@ -18,6 +16,6 @@
...
+  String dateOfCazEntry;
```
plenza: Why String? What's the problem with `LocalDate`?

src/main/java/uk/gov/caz/psr/dto/PaymentStatusRequest.java
plenza: We discussed it with @radekmorytko internally and it seems like due to implementation of validations and bindingresult, there is no better way to do it. So discard my comments.

PR #86, opened 2019-12-09
CAZ-1423 Enforce presence of 'Timestamp' header in charge settlement API

radekmorytko:


src/main/java/uk/gov/caz/psr/dto/GenericErrorResponse.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.caz.psr.dto;
+
+import lombok.Builder;
+import lombok.Value;
+import org.springframework.http.HttpStatus;
+
+@Value
+@Builder
+public class GenericErrorResponse {
```
plenza: At least some super short JDoc please

src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
```diff
@@ -40,7 +41,7 @@
 
   @Override
   public ResponseEntity<PaymentInfoResponse> getPaymentInfo(PaymentInfoRequest request,
-      BindingResult bindingResult, UUID cleanAirZoneId) {
+      BindingResult bindingResult, UUID cleanAirZoneId, LocalDateTime timestamp) {
```
plenza: I'm still wondering what the heck we should do with this? Log? Put to MDC? Discard as you have here? Pretty useless...
Let's leave it as that for now.

PR #89, opened 2019-12-09
CAZ-1584 Add integration tests for `payment-info` endpoint

radekmorytko:
* add integration tests for `payment-info` endpoint
* log payment-info request upon the number of entries returned by the endpoint


src/it/java/uk/gov/caz/psr/ChargeSettlementPaymentInfoTestIT.java
```diff
@@ -0,0 +1,415 @@
...
+    executeSqlFrom("data/sql/clear-all-payments.sql");
+  }
+
+  private void executeSqlFrom(String classPathFile) {
+    ResourceDatabasePopulator populator = new ResourceDatabasePopulator();
+    populator.addScripts(new ClassPathResource(classPathFile));
+    populator.execute(dataSource);
+  }
+
+  private static class PaymentInfoAssertion {
```
plenza: I'd move this class and above helper private method to the bottom of this file. So when me or somebody else opens file to read tests they are close to the top

radekmorytko: Agree, thanks. Fixed

PR #90, opened 2019-12-10
CAZ-1423 Add a null-check for paymentSubmittedTimestamp when returnin…

radekmorytko:
…g data for payment-info endpoint

`paymentSubmittedTimestamp` is set to a non-null value once the call to gov uk pay is successfully completed. Otherwise it is set to null. The previous logic did not take this into account, hence the need to introduce this fix.


src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
```diff
@@ -92,7 +92,6 @@ public Payment create(Payment payment, String returnUrl) {
           .externalId(responseBody.getPaymentId())
           .submittedTimestamp(LocalDateTime.now())
           .externalPaymentStatus(externalPaymentStatus)
-          .submittedTimestamp(LocalDateTime.now())
```
radekmorytko: FYI I removed a duplicated line as on 92 the same assigning happens.

PR #96, opened 2019-12-12
CAZ-1627 Filter out unsuccessful payments from payment-info endpoint

radekmorytko:
* add a new specification that does not include not paid payments
* modify integration tests

### DIFF
--- a/src/it/java/uk/gov/caz/psr/ChargeSettlementPaymentInfoTestIT.java
+++ b/src/it/java/uk/gov/caz/psr/ChargeSettlementPaymentInfoTestIT.java
@@ -63,6 +63,7 @@ class ChargeSettlementPaymentInfoTestIT {
             .then()
             .headerContainsCorrelationId()
             .reponseHasOkStatus()
+            .doesNotContainNotPaidEntries()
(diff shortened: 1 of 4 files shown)


src/it/java/uk/gov/caz/psr/ChargeSettlementPaymentInfoTestIT.java
```diff
@@ -768,6 +783,12 @@ private RequestSpecification commonRequestSpecification() {
           .header(Headers.X_API_KEY, API_KEY_FOR_EXISTING_RECORDS)
           .header(Headers.TIMESTAMP, LocalDateTime.now().toString());
     }
+
+    public PaymentInfoAssertion doesNotContainNotPaidEntries() {
+      validatableResponse.body("results.payments.lineItems.findAll { it.paymentStatus == "
```
plenza: what the hell is this magic?

radekmorytko: Yes, it is possible. As per https://github.com/rest-assured/rest-assured/wiki/usage#jsonpath-configuration
> Note that the JsonPath implementation uses Groovy's GPath syntax and is not to be confused with Jayway's JsonPath implementation.

PR #98, opened 2019-12-12
uniquely generate a message group ID

kingarj:


src/test/java/uk/gov/caz/psr/messaging/MessagingClientTest.java
```diff
@@ -35,7 +35,7 @@
 
   @BeforeEach
   void init() throws JsonProcessingException {
-    messagingClient = new MessagingClient("testId", "testQueue", client, new ObjectMapper());
+    messagingClient = new MessagingClient("testQueue", client, new ObjectMapper());
     sendEmailRequest = SendEmailRequest.builder().emailAddress("test@test.com")
         .personalisation("{}").templateId("test-template-id").build();
     ReflectionTestUtils.setField(messagingClient, "newQueueName", "new");
```
radekmorytko: I think this line is not needed anymore and can be safely removed.

src/main/java/uk/gov/caz/psr/messaging/MessagingClient.java
```diff
@@ -18,27 +18,20 @@
 @Slf4j
 public class MessagingClient {
 
-  @Value("${services.sqs.new-queue-name}")
   String newQueueName;
```
radekmorytko: This field can be `private final` as well.

PR #99, opened 2019-12-13
Change reference to secret in properties

kingarj:
DEV environment not currently finding the secret containing the API keys (as it currently uses the bootstrap properties). I think the best fix for this is to move away from the bootstrap properties (as they are used to populate properties from aws secrets - these just happen to be stored together) and store the name of the secret separately.
This mimics the pattern we are using on Vehicle Checker.


src/main/resources/application.yml
```diff
@@ -41,6 +41,10 @@ management:
     diskspace:
       enabled: false
 
+
+aws:
+  secret-name: /config/payments.dev
```
radekmorytko: Will this value be the same for environments other than `dev`?

src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
```diff
@@ -189,7 +189,7 @@ private URI buildFindByUri(String id) {
   private RequestEntity<CreateCardPaymentRequest> buildRequestEntityForCreate(String apiKey,
       CreateCardPaymentRequest body) {
     return RequestEntity.post(URI.create(rootUrl + CREATE_URI))
-        .header("Authorization", "Bearer " + apiKey).accept(MediaType.APPLICATION_JSON)
+        .header("Authorization", "Bearer" + apiKey).accept(MediaType.APPLICATION_JSON)
```
radekmorytko: I can see the space has been removed. Does the `apiKey` format now include the space?

src/main/resources/application.yml
kingarj: It won't but I will ask Lakshman for a new environment variable that can update this for SIT and subsequent environments.

src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
kingarj: Yes, when I tested it locally for some reason there were the two spaces - at some point there must be a space tacked on to the start so taking this one out creates a token with the correct format.

radekmorytko: I don't know know whether you tested using AWS or localstack, but please be cautious when it comes to testing with `localstack` - sometimes there are subtle yet important differences between it and AWS which causes a lot of trouble.

If you were testing with localstack, could you please verify the same behaviour for AWS? Or,
maybe the 'safest' solution (which doesn't take care of whitespaces) would be to trim `apiKey` before the usage (e.g. by [trimWhitespace](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html#trimWhitespace-java.lang.String-)) and leave the space as it was before?

src/main/resources/application.yml
radekmorytko: OK, in that case, could you please add a comment that this value will be overridden externally for envs other than `dev`?
```
aws:
  secret-name: /config/payments.dev # for other envs will be externally overridden
```

src/main/java/uk/gov/caz/psr/repository/ExternalPaymentsRepository.java
kingarj: Agreed with the 'safest' solution - have added the trim method call.

PR #109, opened 2020-01-10
CAZ-1730 added logging payment request

jarnie86:


src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
```diff
@@ -9,10 +9,12 @@
 import javax.validation.constraints.Positive;
 import javax.validation.constraints.Size;
 import lombok.Builder;
+import lombok.ToString;
 import lombok.Value;
 
 @Value
 @Builder
+@ToString
```
plenza: `@Value` already adds `@ToString`

PR #113, opened 2020-01-14
[do not merge] Generate central reference number and return to front end

kingarj:
- Creates a new sequence in the public schema which increments, from a seed value, a reference in the Payment table.
- Sends this reference through to Gov.UK Pay
- Returns the Gov.UK Pay ID and the central reference number from the initiate payment endpoint
- Adds parameters needed to build new Notify template for payments receipt
- Adds reference to the response received from GET requests to the Charge Settlement API.

**Needs to be merged in tandem with corresponding PR in the Payments UI**


cruddasj:
@plenza - FYI for visibility on the above. This won't be merged directly into develop, rather it is simply highlighting the code changes that will be layered on top of the new ERD branch currently being worked on.

cruddasj:
Please do not merge this directly into develop. The merge should be made into the new ERD branch then a single mergedown to develop.

PR #114, opened 2020-01-15
[CAZ-1720] Finish Payment with new ERD

dratka:
Jira Card: https://eaflood.atlassian.net/browse/CAZ-1720
Updated all repositories and services which handle Finish payment after payment. 
I decided to rename CazEntrantPayment to EntrantPayment with all related classes.


src/main/java/uk/gov/caz/psr/service/PaymentUpdateStatusBuilder.java
```diff
@@ -2,23 +2,29 @@
...
 import org.springframework.stereotype.Service;
+import uk.gov.caz.psr.model.EntrantPayment;
+import uk.gov.caz.psr.model.EntrantPaymentUpdateActor;
 import uk.gov.caz.psr.model.ExternalPaymentDetails;
 import uk.gov.caz.psr.model.ExternalPaymentStatus;
+import uk.gov.caz.psr.model.InternalPaymentStatus;
 import uk.gov.caz.psr.model.Payment;
 
 /**
  * Creates a new instance of {@link Payment} with the new external status set and a mapped internal
```
radekmorytko: Could you please update this javadoc and the ones for private methods too? It is essential to justify the logic on line 63, i.e. creation of an empty array when status is equal to `InternalPaymentStatus.PAID`

src/main/java/uk/gov/caz/psr/service/PaymentUpdateStatusBuilder.java
```diff
@@ -48,20 +52,24 @@ private void checkPreconditions(Payment payment, ExternalPaymentDetails external
...
-  //  }
+  /**
+   * Creates a new list of {@link EntrantPayment} with an internal status mapped from {@code
+   * status}.
+   */
+  private List<EntrantPayment> buildEntrantPaymentsWith(
+      InternalPaymentStatus status,
+      List<EntrantPayment> entrantPayments) {
+    if (!status.equals(InternalPaymentStatus.PAID)) {
+      return new ArrayList<>();
```
radekmorytko: I think we don't need to return a mutable version of an array list - `Collections.emptyList()` should suffice

PR #116, opened 2020-01-16
Move audit table to new schema

kingarj:
Creates a new schema for the audit table and adds triggers to all tables in caz_payment to write to the audit table on insert, update or delete.

### DIFF
--- a/src/main/resources/db/changelog/changes/0002-1.0-move-audit-table-to-new-schema.yaml
+++ b/src/main/resources/db/changelog/changes/0002-1.0-move-audit-table-to-new-schema.yaml
@@ -1,0 +1,15 @@
+databaseChangeLog:
+  - changeSet:
+      id: 0002-1.0
+      author: informed
+      preConditions:
+        - onFail: MARK_RAN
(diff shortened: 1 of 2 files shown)


src/main/resources/db/changelog/changes/0002-1.0-move-audit-table-to-new-schema.yaml
```diff
@@ -0,0 +1,15 @@
+databaseChangeLog:
+  - changeSet:
```
cruddasj: Triggers are all fine. Can you check the script ID to ensure we avoid collision?

kingarj: Yes, the previous changeSet IDs are 0001.1 and 0001.2.

cruddasj: Great 👍

cruddasj:
@plenza - FYI on the above for visibility.

PR #118, opened 2020-01-16
Generate and propagate central reference number

kingarj:
- Creates a new sequence in the public schema which increments, from a seed value, a reference in the Payment table.
- Sends this reference through to Gov.UK Pay
- Returns the Gov.UK Pay ID and the central reference number from the initiate payment endpoint
- Adds parameters needed to build new Notify template for payments receipt
- Adds reference to the response received from GET requests to the Charge Settlement API.

**Needs to be merged in tandem with corresponding PR in the Payments UI**


cruddasj:
@plenza - tagged in for visibility 👍

src/it/java/uk/gov/caz/psr/ReconcilePaymentStatusTestIT.java
```diff
@@ -29,7 +29,7 @@
     executionPhase = ExecutionPhase.AFTER_TEST_METHOD)
 @IntegrationTest
 @AutoConfigureMockMvc
-public class GetAndUpdatePaymentStatusTestIT {
+public class ReconcilePaymentStatusTestIT {
```
radekmorytko: Unfortunately this change is not sufficient - tests need to be adjusted as well since currently they make `get` requests instead of `put`.

src/it/java/uk/gov/caz/psr/GetChargeSettlementPaymentStatusTestIT.java
```diff
@@ -115,7 +119,7 @@ public void shouldReturn200AndTheExistingPaymentWhenNoPaidEntrantExists(String v
         .andExpect(status().isOk())
         .andExpect(content().json(
             getResponseWith(InternalPaymentStatus.NOT_PAID, VALID_CASE_REFERENCE,
-                VALID_EXTERNAL_ID_FOR_NOT_PAID)));
+                VALID_EXTERNAL_ID_FOR_NOT_PAID, VALID_PAYMENT_REFERENCE)));
```
radekmorytko: When I run this test it fails (I didn't look into its specifics):
```
java.lang.AssertionError: cazPaymentReference
Expected: 3001
     got: 3000
```

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -26,6 +26,7 @@
       + "entrant_payment.payment_status, "
       + "entrant_payment.case_reference, "
       + "payment.payment_provider_id "
```
radekmorytko: Please add `,` at the end of the string - with the new column the query's syntax is wrong.

src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
```diff
@@ -48,6 +48,7 @@
     private static final String PAYMENT_METHOD = "payment_method";
     private static final String TOTAL_PAID = "total_paid";
     private static final String PAYMENT_PROVIDER_STATUS = "payment_provider_status";
+    private static final String REFERENCE_NUMBER = "central_reference_number";
```
radekmorytko: In `uk.gov.caz.psr.repository.PaymentRepository.Sql#FIND_BY_ENTRANT_PAYMENT_ID_SQL` please add this column (`payment.central_reference_number, `) to the query as it causes errors in integration tests.

PR #121, opened 2020-01-17
[CAZ-1715] Sync payments - payments does not exist

musztardem:


src/main/java/uk/gov/caz/psr/service/EntrantPaymentService.java
```diff
@@ -25,30 +26,60 @@
...
-        // TODO: CAZ-1715
-      }
+      Optional<EntrantPayment> record = fetchEntrantPaymentFromRepository(vehicleEntrantDto);
+
+      EntrantPayment entrantPayment = record.isPresent()
+          ? record.get()
+          : storeNewEntrantPayment(vehicleEntrantDto);
+
+      updateVehicleEntrantCapturedIfNotCaptured(entrantPayment);
+      addEntrantPaymentDtoToReturnCollection(entrantPaymentsDtosReturnCollection,
```
radekmorytko: My own preference is to avoid methods that modify the passed arguments and returns nothing - it is not a big deal, but I think it simplifies the reasoning about the logic. What do you think about 
- renaming `entrantPaymentsDtosReturnCollection` to `result`
- inlining the whole method. Line `39` would look as follows:
```java
result.add(EntrantPaymentDto.from(entrantPayment));
```

PR #125, opened 2020-01-20
[CAZ-1724] ChargeSettelment PUT update payment status

musztardem:
[CAZ-1724 JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1724)

Screenshots:

Screenshot 2020-01-20 at 16 59 48
Screenshot 2020-01-20 at 16 59 30

---
_**UPDATE**_
* Changed `chargeSettlementPaymentStatus` to `paymentStatus`: 
* [CAZ-1814 JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1814)
Screenshot 2020-01-21 at 11 11 30


cruddasj:
@musztardem - as part of upcoming works, the payment status field will need renamed to align with the ICD specification (a bug captured under https://eaflood.atlassian.net/browse/CAZ-1814). If this is going through PR review, it might be worth updating the field so we can cover this in one review process? Appreciate the commit trail may be less clean but it would be helpful in terms of moving things along (as it's a quick refactor).

src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdateService.java
```diff
@@ -18,82 +23,84 @@
...
+  }
+
+  private void handleNewEntrantPayment() {
+    // TODO: Implement in CAZ-1723
+  }
+
+  private void handleUpdateEntrantPayment(EntrantPayment entrantPayment,
+      EntrantPaymentStatusUpdate entrantPaymentStatusUpdate) {
+    Optional<Payment> payment = paymentRepository
+        .findByEntrantPayment(entrantPayment.getCleanAirZoneEntrantPaymentId());
```
plenza: @radekmorytko - we may need to talk about that.
Should we select "latest = true" Payment by using "Match" joining table here.. or not?

plenza: Answering myself: this is already implemented in `PaymentRepository.findByEntrantPayment` method

src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdateService.java
```diff
@@ -18,82 +23,84 @@
...
+
+    if (!payment.isPresent()) {
```
radekmorytko: @musztardem I'm wondering what the expected logic should be when a) an entrant is created using this PUT endpoint (entrant payment and payment does not exist), b) there is a request to update it using this endpoint. Currently we're throwing `MissingPaymentForEntrantPaymentException`, because payment is not found. Is this correct?

plenza: In my opinion this PUT aka update-by-LA operation **when PaymentEntrant and Payment exist** (as this is topic of this Jira task) should only check if latest Payment status is "in progress" and if so, throw an Exception.

If Payment is missing, then https://eaflood.atlassian.net/browse/CAZ-1723 should provide proper implementation.

musztardem: @plenza We discussed it shortly with @radekmorytko and we came to the conclusion that it should work the following way:
- When `EntrantPayment` exists and it has no associated payment OR it has associated payment with a status other than `in progress` - update it - this scenario would be implemented in this card.
- When `EntrantPayment` has associated payment and the payment has the `in progress` status - throw an exception (CAZ-1725)
- When `EntrantPayment` is missing - just create it (CAZ-1723)

Therefore, if the exception throwing will be removed from this place, I think the linked CAZ-1858 card won't be necessary. How do you think?

plenza: Conclusion is correct. However I would keep 1858 and track necessary changes as separate task. However, in this case you can easily downgrade 1858 estimation to 1 or 2 I guess as this will probably be 1-liner in prod code. It may/will require additonal tests thou, so I leave final decision to you. If you won't go with 1858 let me know and I'll rename it to mark it for deletion. Thanks!

PR #127, opened 2020-01-22
CAZ-1726 Update charge settlement GET payment-info endpoint to match …

radekmorytko:
…the new ERD


src/it/java/uk/gov/caz/psr/ChargeSettlementPaymentInfoTestIT.java
```diff
@@ -843,7 +841,7 @@ public PaymentInfoAssertion containsEmptyResults() {
     public PaymentInfoAssertion totalLineItemsCountIsEqualTo(int lineItemsCount) {
       validatableResponse
           .body("results.collect { it.payments.collect { it.lineItems.size() }.sum() }.sum()",
-              equalTo(lineItemsCount));
+              equalTo(lineItemsCount == 0 ? null : lineItemsCount));
```
plenza: This line has lost readability. I would extract groovy string into named constant stating what it does (total sum of collective line items).

src/main/java/uk/gov/caz/psr/service/paymentinfo/OmitNotPaidPaymentInfoSpecification.java
```diff
@@ -2,23 +2,27 @@
...
+import uk.gov.caz.psr.model.info.EntrantPaymentInfo_;
+import uk.gov.caz.psr.model.info.EntrantPaymentMatchInfo;
+import uk.gov.caz.psr.model.info.EntrantPaymentMatchInfo_;
 
 /**
  * A specification that does not include vehicle entrant payments with {@code notPaid} status.
  */
-public class OmitNotPaidPaymentInfoSpecification
-    implements Specification<VehicleEntrantPaymentInfo> {
+public class OmitNotPaidPaymentInfoSpecification implements Specification<EntrantPaymentMatchInfo> {
```
plenza: Is there a way to create positive match specification? I mean to have spec that matches paid payments and use it in `ChargeSettlementPaymentInfoService` with `.not` instead of `.and`?
This is minor, because I'm not sure if my proposition is really a better one

src/it/java/uk/gov/caz/psr/ChargeSettlementPaymentInfoTestIT.java
radekmorytko: Good suggestions, thanks

src/main/java/uk/gov/caz/psr/service/paymentinfo/OmitNotPaidPaymentInfoSpecification.java
radekmorytko: I'm not sure if I got your suggestion right, but here's my proposal:
We're supposed to return payments whose status is not equal to `NOT_PAID` (paid, refunded or charged back), therefore creating a specification that finds paid payments and negating it includes payments with {not paid, refunded, charged back} statuses.
What we can do if we want to use `.not` is to rename `OmitNotPaidPaymentInfoSpecification` to `NotPaidPaymentInfoSpecification` with the following condition:
```java
return criteriaBuilder.equal(entrantPaymentInfoJoin.get(EntrantPaymentInfo_.paymentStatus), InternalPaymentStatus.NOT_PAID);
```
Then, in `ChargeSettlementPaymentInfoService`, we can use it with `.not`:
```java
private static final NotPaidPaymentInfoSpecification NOT_PAID_PAYMENT_INFO_SPECIFICATION
    = new NotPaidPaymentInfoSpecification();
...
private Specification<EntrantPaymentMatchInfo> initialSpecification(UUID cazId) {
  return CazIdSpecification.forCaz(cazId).and(Specification.not(NOT_PAID_PAYMENT_INFO_SPECIFICATION));
}
```
What do you think?

plenza: Thanks for example. I like current approach better, so skip my suggestion :)

PR #130, opened 2020-01-23
Add central reference and external id to initiate payment response

kingarj:


kingarj:
Made redundant through more appropriate refactoring of UI.

PR #131, opened 2020-01-24
[CAZ-1727] Fix Dangling Payments services for new ERD

dratka:
Jira card: https://eaflood.atlassian.net/browse/CAZ-1727
Updated services with new ERD for Dangling Payment.

### DIFF
--- a/src/it/java/uk/gov/caz/psr/DanglingPaymentsCleanupTestIT.java
+++ b/src/it/java/uk/gov/caz/psr/DanglingPaymentsCleanupTestIT.java
@@ -1,260 +1,290 @@
-//package uk.gov.caz.psr;
-//
-//import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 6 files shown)


radekmorytko:
I think we need to handle in a more 'robust' way the extreme edge case: a payment is finished successfully on the GOV UK Pay's side, but the customer didn't visit our UI and the finalisation process is done via the dangling payment mechanism and an email is sent to the customer. This is the case for the email:
Currently we should send CAZ name in the emai. It is passed to us by UI. However, when we update/finalise the payment by the dangling payment handler, we do not have this information and return `null`. This happens here: `uk/gov/caz/psr/service/listener/PaymentReceiptSender.java:41`:
```java
String cazName = payment.getCleanAirZoneName();
```
Could you please extract this bit to a private method, log when it is `null` with the `WARN`, return a more friendly name (e.g. "CAZ with ID #{caz_id}") and add a unit test?

When I was asked to consider the solution for this problem I have forgotten about this edge case.

PR #133, opened 2020-01-27
[CAZ-1725] PUT /payment-status - pending payment exception

musztardem:
[CAZ-1725 JIRA Card](https://eaflood.atlassian.net/browse/CAZ-1725)

### DIFF
--- a/src/it/java/uk/gov/caz/psr/ErrorPaymentStatusUpdateTestIT.java
+++ b/src/it/java/uk/gov/caz/psr/ErrorPaymentStatusUpdateTestIT.java
@@ -39,6 +39,7 @@ import org.springframework.test.jdbc.JdbcTestUtils;
 import uk.gov.caz.correlationid.Constants;
 import uk.gov.caz.psr.annotation.FullyRunningServerIntegrationTest;
 import uk.gov.caz.psr.controller.ChargeSettlementController;
+import uk.gov.caz.psr.dto.ChargeSettlementPaymentStatus;
 import uk.gov.caz.psr.dto.Headers;
 import uk.gov.caz.psr.dto.PaymentStatusUpdateDetails;
 import uk.gov.caz.psr.dto.PaymentStatusUpdateRequest;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdateService.java
```diff
@@ -58,13 +59,9 @@ private void handleNewEntrantPayment(EntrantPaymentStatusUpdate entrantPaymentSt
 
   private void handleUpdateEntrantPayment(EntrantPayment entrantPayment,
       EntrantPaymentStatusUpdate entrantPaymentStatusUpdate) {
-    Optional<Payment> payment = paymentRepository
+    Optional<Payment> optionalPayment = paymentRepository
```
radekmorytko: I would slightly refactor this code so that it is obvious to the reader what we are checking/verifying here. For example, lines 62-64 can be merged to one method call:
```java
private void handleUpdateEntrantPayment(EntrantPayment entrantPayment,
    EntrantPaymentStatusUpdate entrantPaymentStatusUpdate) {
  verifyAssociatedPaymentIsFinished(entrantPayment);
  ...
}
...
private void verifyAssociatedPaymentIsFinished(EntrantPayment entrantPayment) {
  Payment payment = paymentRepository
      .findByEntrantPayment(entrantPayment.getCleanAirZoneEntrantPaymentId())
      .orElse(null);
  if (payment != null && payment.getExternalPaymentStatus().isNotFinished()) {
    log.error("Payment with ID: {} is still being processed.", payment.getId());
    throw new PaymentNotProcessedException("Payment is still being processed");
  }
}
```

src/it/java/uk/gov/caz/psr/ErrorPaymentStatusUpdateTestIT.java
```diff
@@ -144,6 +165,35 @@ public PaymentStatusUpdateJourneyAssertion whenInvalidRequestSubmitted() {
...
+          .contentType(MediaType.APPLICATION_JSON.toString())
+          .header(Constants.X_CORRELATION_ID_HEADER, correlationId)
+          .header(Headers.TIMESTAMP, LocalDateTime.now().toString())
+          .header(Headers.X_API_KEY, cleanAirZoneId)
+          .body(toJsonString(paymentStatusUpdateRequest))
+          .when()
+          .put()
+          .then()
+          .header(Constants.X_CORRELATION_ID_HEADER, correlationId)
+          .header(STRICT_TRANSPORT_SECURITY_HEADER, STRICT_TRANSPORT_SECURITY_VALUE)
```
radekmorytko: There are other tests that take care of headers on lines 184-189, I think we don't need to duplicate it here. E.g. making sure that `X_CORRELATION_ID_HEADER` is present should be enough :)

PR #143, opened 2020-01-29
[CAZ-1870] Return 404 for PaymentStatus

dratka:
Jira card: https://eaflood.atlassian.net/browse/CAZ-1870

Updated the process to return 404 when EntrantPayment not found. 
Added case to return 404 when EntrantPayment not captured in VCCS and failed payment has been processed. 
Added new test cases.

### DIFF
--- a/src/it/java/uk/gov/caz/psr/SuccessPaymentStatusUpdateTestIT.java
+++ b/src/it/java/uk/gov/caz/psr/SuccessPaymentStatusUpdateTestIT.java
@@ -81,7 +81,7 @@ public class SuccessPaymentStatusUpdateTestIT {
         .paymentStatusUpdateRequest(paymentStatusUpdateRequest(vrn))
         .whenSubmitted()
         .then()
-        .entrantPaymentsAreUpdatedInTheDatabse();
+        .entrantPaymentsAreUpdatedInTheDatabase();
   }
 
   @Test
@@ -125,6 +125,7 @@ public class SuccessPaymentStatusUpdateTestIT {
     private final ObjectMapper objectMapper;
     private final JdbcTemplate jdbcTemplate;
     private final UUID cleanAirZoneId = UUID.fromString("b8e53786-c5ca-426a-a701-b14ee74857d4");
+    private final int EXPECTED_NUMBER_OF_REFUNDED_RECORDS = 4;
 
     private PaymentStatusUpdateRequest paymentStatusUpdateRequest;
 
@@ -169,7 +170,7 @@ public class SuccessPaymentStatusUpdateTestIT {
       return objectMapper.writeValueAsString(request);
     }
 
-    public PaymentStatusUpdateJourneyAssertion entrantPaymentsAreUpdatedInTheDatabse() {
+    public PaymentStatusUpdateJourneyAssertion entrantPaymentsAreUpdatedInTheDatabase() {
       verifyThatRefundedEntrantPaymentsExists();
       return this;
     }
@@ -185,7 +186,7 @@ public class SuccessPaymentStatusUpdateTestIT {
           "clean_air_zone_id = '" + cleanAirZoneId.toString() + "' AND "
               + "vrn = 'ND84VSX' AND "
               + "payment_status = '" + InternalPaymentStatus.REFUNDED.name() + "'");
-      assertThat(vehicleEntrantCount).isEqualTo(3);
+      assertThat(vehicleEntrantCount).isEqualTo(EXPECTED_NUMBER_OF_REFUNDED_RECORDS);
     }
 
     private void verifyThatNewEntrantPaymentWasCreated() {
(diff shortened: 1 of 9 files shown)


src/it/java/uk/gov/caz/psr/SuccessPaymentStatusUpdateTestIT.java
```diff
@@ -185,7 +185,7 @@ private void verifyThatRefundedEntrantPaymentsExists() {
           "clean_air_zone_id = '" + cleanAirZoneId.toString() + "' AND "
               + "vrn = 'ND84VSX' AND "
               + "payment_status = '" + InternalPaymentStatus.REFUNDED.name() + "'");
-      assertThat(vehicleEntrantCount).isEqualTo(3);
+      assertThat(vehicleEntrantCount).isEqualTo(4);
```
plenza: Would be good to have this 3(4) extracted to CONSTANT and named appropriately

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -28,15 +28,20 @@
...
+      + "LEFT OUTER JOIN caz_payment.t_payment payment "
       + "ON entrant_payment_match.payment_id = payment.payment_id "
       + "WHERE entrant_payment.clean_air_zone_id = ? AND "
       + "entrant_payment.vrn = ? AND "
-      + "entrant_payment.travel_date = ?";
+      + "entrant_payment.travel_date = ? AND "
+      // exclude not captured with failed payment records.
+      + "(entrant_payment.vehicle_entrant_captured is true OR "
+      + "entrant_payment.vehicle_entrant_captured is false AND "
+      + "entrant_payment.update_actor != 'User' AND "
```
radekmorytko: Are we sure that `"User"` is the correct value? I'm asking, because all values of the enum are uppercase:
```java
public enum EntrantPaymentUpdateActor {
  USER,
  VCCS_API,
  LA
}
```
I guess `'USER'` should be used.

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -28,15 +28,20 @@
...
       + "entrant_payment_match.clean_air_zone_entrant_payment_id "
       + "AND entrant_payment_match.latest = true "
-      + "INNER JOIN caz_payment.t_payment payment "
+      + "LEFT OUTER JOIN caz_payment.t_payment payment "
       + "ON entrant_payment_match.payment_id = payment.payment_id "
       + "WHERE entrant_payment.clean_air_zone_id = ? AND "
       + "entrant_payment.vrn = ? AND "
-      + "entrant_payment.travel_date = ?";
+      + "entrant_payment.travel_date = ? AND "
+      // exclude not captured with failed payment records.
```
radekmorytko: I'm not sure whether this query is correct. In the spec we're told that we should exclude rows whose
> status are set to *notPaid* and vehicle_entrant_captured set to *false* and update actor set to *User*

In SQL it is mapped to:
```sql
{BASE_QUERY}
and status = 'NOT_PAID'
and vehicle_entrant_captured is false
and update_actor = 'USER'
```
Therefore, the only records we *must include* are the ones which satisfy the *negated* aforementioned condition:

```sql
{BASE_QUERY}
and (status != 'NOT_PAID'
or vehicle_entrant_captured is true
or update_actor != 'USER')
```

Whereas currently the condition we have lacks `status != 'NOT_PAID'` part:
```sql
{BASE_QUERY}
and (entrant_payment.vehicle_entrant_captured is true OR
-- I omit "entrant_payment.vehicle_entrant_captured is false" because in this 'or' clause this is implicit
entrant_payment.update_actor != 'USER')
```

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -28,15 +29,19 @@
...
+      + "(entrant_payment.vehicle_entrant_captured is true OR "
+      + "entrant_payment.update_actor != '" + EntrantPaymentUpdateActor.USER + "' OR "
```
plenza: By writing `EntrantPaymentUpdateActor.USER` you are really using `toString()` method while I think you want `.name()` method here.

src/test/java/uk/gov/caz/psr/service/ChargeSettlementServiceTest.java
```diff
@@ -87,10 +88,10 @@ void shouldReturnTheExistingPaymentStatusWhenThereIsChargebackOne() {
     mockSingleEntrantPaymentForStatus(InternalPaymentStatus.CHARGEBACK);
 
     //when
-    PaymentStatus paymentStatus = chargeSettlementService
+    Optional<PaymentStatus> paymentStatus = chargeSettlementService
         .findChargeSettlement(ANY_CAZ_ID, ANY_VRN, ANY_DATE);
 
-    assertThat(paymentStatus.getStatus()).isEqualTo(InternalPaymentStatus.CHARGEBACK);
+    assertThat(paymentStatus.get().getStatus()).isEqualTo(InternalPaymentStatus.CHARGEBACK);
```
radekmorytko: Instead of using not-so-pretty `get` you can use the following:
```java
assertThat(paymentStatus).map(PaymentStatus::getStatus).contains(...)
```

PR #146, opened 2020-01-30
[CAZ-1904] Return 400 for non-existant vehicle passage

kingarj:
As per ticket [CAZ-1904](https://eaflood.atlassian.net/browse/CAZ-1904), if no record of the vehicle passage exists then a 400 error response should be returned.


src/main/java/uk/gov/caz/psr/service/PaymentStatusUpdateService.java
```diff
@@ -46,17 +45,12 @@ public void process(
       if (entrantPayment.isPresent()) {
         handleUpdateEntrantPayment(entrantPayment.get(), entrantPaymentStatusUpdate);
       } else {
-        handleNewEntrantPayment(entrantPaymentStatusUpdate);
+        log.error("Vehicle entry not found");
+        throw new PaymentDoesNotExistException(entrantPaymentStatusUpdate.getVrn());
```
dratka: You could refactor this condition a bit to use .orElseThrow because entrantPayment is the object of Optional class.

src/main/java/uk/gov/caz/psr/service/exception/PaymentDoesNotExistException.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import uk.gov.caz.ApplicationRuntimeException;
+
+/**
+ * Exception class which will be used to throw exception when {@link
+ * uk.gov.caz.psr.model.Payment} is supposed to be updated but
+ * the payment is still being processed.
+ */
+@ResponseStatus(value = HttpStatus.BAD_REQUEST)
```
dratka: For exceptions handled in `ExceptionController` there is no need to set the `ResponseStatus`.

src/main/java/uk/gov/caz/psr/service/exception/PaymentDoesNotExistException.java
```diff
@@ -0,0 +1,23 @@
...
+@Value
+public class PaymentDoesNotExistException extends ApplicationRuntimeException {
```
dratka: Here is no need to use `ApplicationRuntimeException` if you are handling it in custom way in `ExceptionController`.  It can be just `RuntimeException`

kingarj:
Hi @dratka, these should all be resolved now.

cruddasj:
@kingarj - in the morning please can you look to resolve the merge conflict on this branch? @dratka - would appreciate if after the conflicts are resolved we can get a re-review.

PR #147, opened 2020-01-30
CAZ-1871 Support for multiple VRNs and different rates per day in POST /payments

radekmorytko:


src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
```diff
@@ -1,44 +1,136 @@
...
+import uk.gov.caz.psr.controller.exception.InvalidRequestPayloadException;
 
 @Value
 @Builder
+@Slf4j
 public class InitiatePaymentRequest {
 
+  private static final Map<Function<InitiatePaymentRequest, Boolean>, String> validators =
+      ImmutableMap.<Function<InitiatePaymentRequest, Boolean>, String>builder()
+          .put(request -> Objects.nonNull(request.getCleanAirZoneId()),
```
plenza: Any chance to make it more eyes friendly?
For example (code written here without verification):
```
.put(cleanAirZoneIdNotNull(), "message"),
.put(returnUrlNotEmpty(), "message"),
.put(vrnInAllTransactionsNotEmpty(),  "message"),
```

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
```diff
@@ -1,44 +1,136 @@
...
+import org.springframework.util.StringUtils;
+import uk.gov.caz.psr.controller.exception.InvalidRequestPayloadException;
 
 @Value
 @Builder
+@Slf4j
 public class InitiatePaymentRequest {
 
+  private static final Map<Function<InitiatePaymentRequest, Boolean>, String> validators =
+      ImmutableMap.<Function<InitiatePaymentRequest, Boolean>, String>builder()
```
plenza: We should definitely use [MapPreservingOrderBuilder](https://github.com/InformedSolutions/JAQU-CAZ-Accounts-API/blob/develop/src/main/java/uk/gov/caz/accounts/util/MapPreservingOrderBuilder.java) here, for example like [here](https://github.com/InformedSolutions/JAQU-CAZ-Accounts-API/blob/0e5116493abc586dd2c70cda4c41dd2c8622cb7b/src/main/java/uk/gov/caz/accounts/dto/AccountCreationRequestDto.java#L41)

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
```diff
@@ -1,44 +1,136 @@
...
+          .put(request -> !CollectionUtils.isEmpty(request.getTransactions()),
+              "'transactions' cannot be null or empty")
+          .put(request -> allTransactionsMatch(request,
+              transaction -> StringUtils.hasText(transaction.getVrn())),
+              "'vrn' cannot be null or empty")
+          .put(request -> allTransactionsMatch(request,
+              transaction -> transaction.getVrn().length() >= 1
+                  && transaction.getVrn().length() <= 15), "'vrn' length must be from 1 to 15")
+          .put(request -> allTransactionsMatch(request,
+              transaction -> Objects.nonNull(transaction.getCharge())), "'charge' cannot be null")
```
plenza: Do we want (or will want?) to have more precise error messages saying that attribute of `Transaction` with such and such `vrn` is invalid? Probably not?

But maybe I would rephrase error message like this: `'charge' in all transactions cannot be null`

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
```diff
@@ -1,44 +1,136 @@
...
+  @ApiModelProperty(value = "${swagger.model.descriptions.payments-initiate.transactions}")
+  List<Transaction> transactions;
 
-  @ApiModelProperty(value = "${swagger.model.descriptions.payments-initiate.amount}")
-  @NotNull
-  @Positive
-  Integer amount;
+  @Value
+  @Builder(toBuilder = true)
+  public static class Transaction {
```
plenza: Javadoc

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
```diff
@@ -1,44 +1,136 @@
...
+        throw new InvalidRequestPayloadException(message);
+      }
+    });
+  }
+
+  /**
+   * Verifies if {@code request} does not contain duplicated vehicle entrants.
+   */
+  private static boolean containsNoDuplicatedEntrants(InitiatePaymentRequest request) {
+    Map<String, List<LocalDate>> transactionDatesByVrn = groupByVrn(request);
```
plenza: A lot is going on here and it is hard to see what it is. Please try to extract to smaller helpers.

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
radekmorytko: I have double checked the documentation and ImmutableMaps preserve insertion order, so we should be fine using it:
> By default, a Builder will generate maps that iterate over entries in the order they were inserted into the builder, equivalently to LinkedHashMap

source: https://guava.dev/releases/28.0-jre/api/docs/com/google/common/collect/ImmutableMap.Builder.html

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
radekmorytko: Good idea, changed

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
radekmorytko: I changed the validation error messages.

src/main/java/uk/gov/caz/psr/dto/InitiatePaymentRequest.java
radekmorytko: I tried to make it more readable.

PR #150, opened 2020-01-31
[CAZ-1900] Disallowed to use notPaid paymentStatus in update

dratka:
Jira card: https://eaflood.atlassian.net/browse/CAZ-1900

Added validation to DTO 
- it validates if the provided value is in specified possible values. 
- updated the type to String to avoid exception on request params mapping when enum value is not exactly the same as defined.

### DIFF
--- a/src/it/java/uk/gov/caz/psr/ErrorPaymentStatusUpdateTestIT.java
+++ b/src/it/java/uk/gov/caz/psr/ErrorPaymentStatusUpdateTestIT.java
@@ -53,6 +53,8 @@ import uk.gov.caz.psr.util.TestObjectFactory.PaymentStatusUpdateDetailsFactory;
     executionPhase = ExecutionPhase.AFTER_TEST_METHOD)
 public class ErrorPaymentStatusUpdateTestIT {
 
+  private static final String VALID_VRN = "ND84VSX";
+
   @LocalServerPort
   int randomServerPort;
 
@@ -100,6 +102,10 @@ public class ErrorPaymentStatusUpdateTestIT {
         .whenNonExistentRequestSubmitted(NON_EXISTING_VRN)
         .then()
         .noEntrantPaymentUpdatedInDatabase();
+
+    given()
+        .paymentStatusUpdateRequest(paymentStatusUpdateRequestWithInvalidPaymentStatus())
+        .failsWhenInvalidStatusRequestSubmitted();
   }
 
   private PaymentStatusUpdateJourneyAssertion given() {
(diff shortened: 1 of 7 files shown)


src/it/java/uk/gov/caz/psr/ErrorPaymentStatusUpdateTestIT.java
```diff
@@ -165,6 +188,33 @@ public PaymentStatusUpdateJourneyAssertion whenInvalidRequestSubmitted() {
...
+          .contentType(MediaType.APPLICATION_JSON.toString())
+          .header(Constants.X_CORRELATION_ID_HEADER, correlationId)
+          .header(Headers.TIMESTAMP, LocalDateTime.now().toString())
+          .header(Headers.X_API_KEY, cleanAirZoneId)
+          .body(toJsonString(paymentStatusUpdateRequest))
+          .when()
+          .put()
+          .then()
+          .header(Constants.X_CORRELATION_ID_HEADER, correlationId)
+          .header(STRICT_TRANSPORT_SECURITY_HEADER, STRICT_TRANSPORT_SECURITY_VALUE)
```
radekmorytko: The presence of those headers (lines 206-211) is already checked in `whenInvalidRequestSubmitted`, I think you can safely omit them here.

src/main/java/uk/gov/caz/psr/dto/validation/constraint/ValueIn.java
```diff
@@ -0,0 +1,41 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+/**
+ * Custom constraint annotation for checking whether the request contains at least one non-null
```
radekmorytko: I think this is the old-one javadocs :)

src/main/java/uk/gov/caz/psr/dto/validation/constraint/ValueInValidator.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static java.util.stream.Collectors.toSet;
+
+import java.util.Arrays;
+import java.util.Set;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+/**
+ * Custom validator that checks whether the passed argument contains is valid string value.
```
radekmorytko: Please update the javadoc :)

PR #151, opened 2020-01-31
[NON_JIRA] POST /payments/paid - Change empty response format

musztardem:
Screenshot 2020-01-31 at 09 46 33

### DIFF
--- a/src/main/java/uk/gov/caz/psr/service/GetPaidEntrantPaymentsService.java
+++ b/src/main/java/uk/gov/caz/psr/service/GetPaidEntrantPaymentsService.java
@@ -1,10 +1,12 @@
 package uk.gov.caz.psr.service;
 
 import java.time.LocalDate;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/caz/psr/service/GetPaidEntrantPaymentsService.java
```diff
@@ -31,17 +31,15 @@
    * @param cleanAirZoneId CAZ in which the payments are checked.
    * @return Map with VRN as a key and list of {@link EntrantPayment} as value.
    */
-  public Map<String,List<EntrantPayment>> getResults(List<String> vrns, LocalDate startDate,
+  public Map<String, List<EntrantPayment>> getResults(List<String> vrns, LocalDate startDate,
```
radekmorytko: As discussed offline:
- to gracefully handle duplicated VRNs we can change `List` to `Set`
- we can change the logic to more declarative:
```java
public Map<String, List<EntrantPayment>> getResults(Set<String> vrns, LocalDate startDate,
    LocalDate endDate, UUID cleanAirZoneId) {
  Map<String, List<EntrantPayment>> results = vrns.stream()
      .collect(
          Collectors.toMap(
              Function.identity(),
              vrn -> entrantPaymentRepository.findAllPaidByVrnAndDateRangeAndCazId(vrn, startDate,
                  endDate, cleanAirZoneId)
          )
      );
  return results;
}
```

PR #154, opened 2020-02-03
CAZ-1921 Unblock abandoned payments

radekmorytko:
See the description in JIRA: https://eaflood.atlassian.net/browse/CAZ-1921

### DIFF
--- a/src/main/java/uk/gov/caz/psr/service/InitiateEntrantPaymentsService.java
+++ b/src/main/java/uk/gov/caz/psr/service/InitiateEntrantPaymentsService.java
@@ -4,9 +4,11 @@ import static uk.gov.caz.psr.util.AttributesNormaliser.normalizeVrn;
 
 import java.time.LocalDate;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import uk.gov.caz.psr.model.EntrantPayment;
 import uk.gov.caz.psr.model.EntrantPaymentMatch;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/caz/psr/service/InitiateEntrantPaymentsService.java
```diff
@@ -59,23 +63,34 @@ void processEntrantPaymentsForPayment(UUID paymentId, int total, List<LocalDate>
...
         .map(Payment::getExternalPaymentStatus)
         .orElse(null);
 
-    if (paymentStatus != null && paymentStatus.isNotFinished()) {
-      throw new IllegalStateException("The corresponding payment has not finished yet, its state"
-          + " is equal to " + paymentStatus);
+    if (relatedPaymentStatus != null && (relatedPaymentStatus.isNotFinished()
+        || relatedPaymentStatus == ExternalPaymentStatus.SUCCESS)) {
+      throw new IllegalStateException("The corresponding payment has been completed or not "
+          + "finished yet, its state is equal to " + relatedPaymentStatus);
```
plenza: Is it possible now that here we have anything other that "finished" status (success or error)? So no way to have "not finished" error?

radekmorytko: It is still possible to get "not finished" error - it happens when someone does not finish their payment's journey. In such scenario, we have a "dangling payment", where a payment is started, but not in a "finished" state (i.e. it is either initiated, created, started or submitted).
The above logic retrieves its status (in `cleanupDanglingPaymentService::processDanglingPayment`) and an exception is thrown since the predicate `relatedPaymentStatus.isNotFinished()` is true.

plenza: I was under the assumption that line 86: `cleanupDanglingPaymentService::processingDanglingPayment` takes care of such statuses, so in here, lines 90-91 there is no way to not be finished

radekmorytko: Okay, I know where the confusion comes from, I think that the naming is not that good as we are processing a *candidate* for a dangling payment, we're not sure if it is a dangling one. If the payment is indeed a dangling one, then you're right, it cannot be in an 'unfinished' state.

PR #160, opened 2020-02-04
Add travel dates to payment receipt email

kingarj:


cruddasj:
@kingarj - it looks like this is a PMD issue which just needs a cleanup on the unused import

PR #162, opened 2020-02-04
[CAZ-1879] Get CleanAirZoneName from VCCS

dratka:
Jira card: https://eaflood.atlassian.net/browse/CAZ-1879
Added Services and Repositories to handle VCCS call for the CAZ name. 
Restored reverted changes https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/159


src/main/java/uk/gov/caz/psr/dto/CleanAirZonesResponse.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.Value;
+import uk.gov.caz.psr.repository.exception.CleanAirZoneNotFoundException;
+
+/**
+ * A value object that represents a response from VCCS request for CleanAirZones details.
+ */
+@Value
+@Builder
+public class CleanAirZonesResponse {
+  private List<CleanAirZoneDto> cleanAirZones;
```
radekmorytko: You can omit `private`, `@Value` is private by default

src/main/java/uk/gov/caz/psr/configuration/AsyncRestConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+  @Value("${services.vccs.rootUrl}")
+  private String vccsApiEndpoint;
+
+  /**
+   * Retrofit2 ntr spring bean.
+   *
+   * @return {@link Retrofit}
+   */
+  @Bean
+  public Retrofit vccsRetrofit(JacksonConverterFactory jacksonConverterFactory) {
```
radekmorytko: I think we don't need to expose three objects as beans, one is enough. Additionally, we can pass `vccsApiEndpoint` as a method parameter:
```java
@Bean
public VccsRepository vccsRepository(ObjectMapper objectMapper,
    @Value("${services.vccs.rootUrl}") String vccsApiEndpoint) {
  return new Retrofit.Builder()
      .baseUrl(requireNonNull(HttpUrl.parse(formatUrl(vccsApiEndpoint))))
      .addConverterFactory(JacksonConverterFactory.create(objectMapper))
      .build()
      .create(VccsRepository.class);
}
```

src/main/java/uk/gov/caz/psr/dto/CleanAirZoneDto.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.UUID;
+import lombok.Builder;
+import lombok.Value;
+
+/**
+ * A value object that represents a single CleanAirZone details from VCCS request.
+ */
+@Value
+@Builder
+public class CleanAirZoneDto {
```
radekmorytko: What do you think about nesting this class inside `CleanAirZonesResponse` as a `static inner class`? (In a similar fashion `uk.gov.caz.psr.dto.PaymentInfoResponse` is structured and we have all necessary classes in one place)

src/main/java/uk/gov/caz/psr/repository/VccsRepository.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Retrofit2 repository to create a vccs call.
+ */
+public interface VccsRepository {
+  /**
+   * Method to create retrofit2 vccs for cleanAirZones call.
+   *
+   * @return {@link Call}
+   */
+  @Headers("Content-Type: application/json")
```
radekmorytko: I think we should drop `Content-Type` header and replace it with `Accept` - we're making a GET request, so we don't have a request body and don't need to specify its content type. On the other hand, we want VCCS to return a response in a JSON format, hence `Accept` header should be used.

src/main/java/uk/gov/caz/psr/repository/VccsRepository.java
```diff
@@ -0,0 +1,38 @@
...
+  @GET("v1/compliance-checker/clean-air-zones")
+  Call<CleanAirZonesResponse> findCleanAirZones(@Header("X-Correlation-ID") String correlationId);
```
radekmorytko: I can see that we're using this method in `uk.gov.caz.psr.service.CleanAirZoneNameGetterService#fetch` with `UUID.randomUUID().toString()`. However, when used in the context of a HTTP request we should use the current correlation id. My advice is to:
* remove `@Header("X-Correlation-ID") String correlationId` parameter completely
* add `X-Correlation-ID` header dynamically in the request interceptor. This can be configured in `uk.gov.caz.psr.configuration.AsyncRestConfiguration`, for example:
```java
@Bean
public VccsRepository vccsRepository(ObjectMapper objectMapper,
  @Value("${services.vccs.rootUrl}") String vccsApiEndpoint) {
return new Retrofit.Builder()
    .baseUrl(requireNonNull(HttpUrl.parse(formatUrl(vccsApiEndpoint))))
    .addConverterFactory(JacksonConverterFactory.create(objectMapper))
    .client(buildHttpClient())
    .build()
    .create(VccsRepository.class);
}

private OkHttpClient buildHttpClient() {
  return new Builder()
      .addInterceptor(chain -> {
        Request original = chain.request();
        Request withCorrelationIdHeader = original.newBuilder()
            .header(Constants.X_CORRELATION_ID_HEADER, getOrGenerateCorrelationId(
                MdcCorrelationIdInjector.getCurrentValue()))
            .build();
        return chain.proceed(withCorrelationIdHeader);
      })
      .build();
}

private String getOrGenerateCorrelationId(String correlationIdFromRequest) {
  return correlationIdFromRequest == null
      ? UUID.randomUUID().toString()
      : correlationIdFromRequest;
}
```
The above logic handle the case when `correlationIdFromRequest` is `null`. It happens when the logic is invoked not in the HTTP request context, but, for example, in a lambda invocation context. In our case this can be the dangling payments handler.

radekmorytko: Currently we rely on the default parameters for connection and read timeouts. It is a good practice to set those values explicitly.
To do this you can follow the pattern from NTR and set them in `application.yml`: https://github.com/InformedSolutions/JAQU-CAZ-National-Taxi-Register/blob/8490ff10f7e9a90de2179037902b366300598936/src/main/resources/application.yml#L16

In the code you can use them when `OkHttpClient` is created:
```java
private OkHttpClient buildHttpClient(long readTimeoutInSeconds, long connectTimeoutInSeconds) {
  return new Builder()
      .readTimeout(readTimeoutInSeconds, TimeUnit.SECONDS)
      .connectTimeout(connectTimeoutInSeconds, TimeUnit.SECONDS)
...
```

src/main/java/uk/gov/caz/psr/service/CleanAirZoneNameGetterService.java
```diff
@@ -0,0 +1,45 @@
...
+   * @return {@link String} if the CleanAirZone exist
+   * @throws NullPointerException if {@code cleanAirZoneId} is null
+   */
+  public String fetch(UUID cleanAirZoneId) {
+    Preconditions.checkNotNull(cleanAirZoneId, "cleanAirZoneId cannot be null");
+
+    try {
+      log.info("Get all cleanAirZones name: start");
+      Response<CleanAirZonesResponse> cleanAirZonesResponse = vccsRepository
+          .findCleanAirZonesSync(UUID.randomUUID().toString());
```
radekmorytko: You can safely remove `UUID.randomUUID().toString()` as now it is handled by the interceptor here: https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/pull/162/files#diff-78313aba990fe6777562d92127899186R59

pom.xml
```diff
@@ -18,7 +18,7 @@
     <artifactFinalName>${project.artifactId}-${project.version}</artifactFinalName>
 
     <rest-assured.version>4.0.0</rest-assured.version>
-    <internal.libraries.version>2.6-SNAPSHOT</internal.libraries.version>
+    <internal.libraries.version>2.7-SNAPSHOT</internal.libraries.version>
```
plenza: We are now on `2.8-SNAPSHOT` so you may update it (not important but if can be done easily then why not)

src/main/java/uk/gov/caz/psr/repository/exception/CleanAirZoneNotFoundException.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.web.bind.annotation.ResponseStatus;
+import uk.gov.caz.ApplicationRuntimeException;
+
+
+/**
+ * Exception which is thrown when repository is not able to find CleanAirZone in VCCS api.
+ */
+@ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
+public class CleanAirZoneNotFoundException extends ApplicationRuntimeException {
+  public CleanAirZoneNotFoundException() {
```
plenza: Could you please extend this constructor with UUID that was searched for and update error message accordingly? May help in debugging

src/main/java/uk/gov/caz/psr/dto/CleanAirZonesResponse.java
```diff
@@ -0,0 +1,40 @@
...
+  List<CleanAirZoneDto> cleanAirZones;
+
+  /**
+   * Method which finds single CleanAirZone with provided ID from the List.
+   *
+   * @param cleanAirZoneId id of Clean Air Zone which we want to find
+   * @return Found CleanAirZone DTO.
+   * @throws CleanAirZoneNotFoundException if Clean Air Zone was not found.
+   */
+  public CleanAirZoneDto findCleanAirZone(UUID cleanAirZoneId) {
```
plenza: DTO object should not be burdened with domain logic. I would move this function to CleanAirZoneNameGetterService and throw domain exception from there as well.

src/main/java/uk/gov/caz/psr/dto/CleanAirZonesResponse.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Value
+  @Builder
+  public static class CleanAirZoneDto {
+    UUID cleanAirZoneId;
+    String name;
+    URI boundaryUrl;
+  }
+
```
plenza: You can drop the line

src/main/java/uk/gov/caz/psr/service/CleanAirZoneNameGetterService.java
```diff
@@ -0,0 +1,45 @@
...
+public class CleanAirZoneNameGetterService {
+
+  private final VccsRepository vccsRepository;
+
+  /**
+   * Gets name of clean air zone name from vccs.
+   *
+   * @param cleanAirZoneId id of clean air zone
+   * @return {@link String} if the CleanAirZone exist
+   * @throws NullPointerException if {@code cleanAirZoneId} is null
```
plenza: It throws `CleanAirZoneNotFoundException` also which is important to the reader and the caller.

src/main/java/uk/gov/caz/psr/service/CleanAirZoneNameGetterService.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public String fetch(UUID cleanAirZoneId) {
+    Preconditions.checkNotNull(cleanAirZoneId, "cleanAirZoneId cannot be null");
+
+    try {
+      log.info("Get all cleanAirZones name: start");
```
plenza: These log statements are not that important. Maybe change them to .debug level?

src/main/java/uk/gov/caz/psr/service/listener/PaymentReceiptSender.java
```diff
@@ -41,13 +44,13 @@ public void onPaymentStatusUpdated(PaymentStatusUpdatedEvent event) {
 
     log.info("Processing email event for payment with ID: {}", payment.getId());
 
-    String cazName = payment.getCleanAirZoneName();
+    String cazName = cleanAirZoneNameGetterService.fetch(getCleanAirZoneId(payment));
```
plenza: @radekmorytko - we will have `CleanAirZoneNotFoundException` here now. It will produce 404 to the caller, right? And email will not be send? Is this correct behaviour?

src/it/java/uk/gov/caz/psr/SuccessPaymentsJourneyTestIT.java
```diff
@@ -565,6 +566,17 @@ private void andReturnsSuccessStatus() {
             .withBody(readFile("get-payment-response.json")));
   }
 
+  public void mockVccsResponse() {
```
plenza: There is some duplication between this and `ReconcilePaymentStatusTestIT`, maybe you could add some helper componenent or provide common base class?

src/it/java/uk/gov/caz/psr/service/CleanAirZoneNameGetterServiceTestIT.java
```diff
@@ -0,0 +1,85 @@
...
+import org.mockserver.integration.ClientAndServer;
+import org.mockserver.model.Header;
+import org.mockserver.model.HttpRequest;
+import org.mockserver.model.HttpResponse;
+import org.springframework.beans.factory.annotation.Autowired;
+import uk.gov.caz.psr.annotation.IntegrationTest;
+import uk.gov.caz.psr.repository.exception.CleanAirZoneNotFoundException;
+
+@IntegrationTest
+public class CleanAirZoneNameGetterServiceTestIT {
```
plenza: I don't think this integration test in necessary. Higher level tests for example `ReconcilePaymentStatusTestIT` test this VCCS call and you test different cases by unit test. I'd remove this test.

src/main/java/uk/gov/caz/psr/service/listener/PaymentReceiptSender.java
radekmorytko: You're right, current logic is not correct: if `CleanAirZoneNotFoundException` is thrown the whole payment handling on our side is interrupted (as `cleanAirZoneNameGetterService.fetch(getCleanAirZoneId(payment))` is not in the `try .. catch` block) and rolled back.
I think the desired behaviour is to move this call to the `try .. catch` block to catch and log the execption. In that scenario, the whole payment flow is not interrupted, the email is not sent and the exception is logged.
We can double check it with Krzysiek.

PR #163, opened 2020-02-06
Return 400 response if vrn not found

kingarj:
If no record of a VRN is found in the VehicleEntrants table, then the API should return a 400 'bad request' response from the payments/info endpoint, as per the Charge Settlement ICD.


radekmorytko:
Side note: honestly I don't clearly get the requirement - why would we ever return `400` status code, which indicates a  **client error**, for a request which is syntactically and (can be) semantically correct? The fact that we do have any information about a particular *vrn* is not client's fault and it is not ours. It's hard to say it is any error whatsoever.

src/main/java/uk/gov/caz/psr/repository/EntrantPaymentRepository.java
```diff
@@ -342,6 +351,24 @@ public void update(EntrantPayment entrantPayment) {
...
   }
 
+  /**
+   * Finds number of instances of a VRN in the EntrantPayment table.
+   *
+   * @param cleanAirZoneId provided Clean Air Zone ID.
+   * @param vrn provided VRN number
+   * @return integer
+   */
+  public Integer findOneByVrnAndCaz(UUID cleanAirZoneId, String vrn) {
```
radekmorytko: I think that `countByCazAndVrn` would be a better name since we do not find and return any entrant payments, we're just counting them.

src/main/java/uk/gov/caz/psr/controller/ChargeSettlementController.java
```diff
@@ -53,6 +53,7 @@
       throw new PaymentInfoDtoValidationException("paymentInfo.validationErrorTitle",
           bindingResult);
     }
+    chargeSettlementPaymentInfoService.throwIfNonExistentVrn(cleanAirZoneId, request.getVrn());
```
radekmorytko: What do you think about having all business logic in services? In this case:
* making `throwIfNonExistentVrn` private
* calling `throwIfNonExistentVrn` as the the first thing in `findPaymentInfo`

kingarj:
@radekmorytko I've addressed your two points of feedback - many thanks. As for your point on the requirement, I believe this may come back to the business case of why the end user in question would make use of this endpoint, which I believe is predominantly for case management, hence there wouldn't be a reason for them to supply a VRN completely unrecognised by the system.
In addition, this specific use case is called out in the ICD so I think this behaviour will be expected by the end user.

PR #164, opened 2020-02-13
Return 400 response if vrn not found (#163)

cruddasj:
Update ST with 404 error response in Charge Settlement API


cruddasj:
@kingarj - as a brief note, this brings in the 400 response in the case a VRN cannot be found under the v1.7 Charge Settlement API.

cruddasj:
Drone run all good 👍

PR #165, opened 2020-02-14
[CAZ-1958] Vehicle retrieval happy path

kingarj:
Please note that this Pull Request encompasses only "happy path" functionality to retrieve and check compliance of pages of vehicles. This is to facilitate development of the front-end page; edge and error case development for this feature will continue concurrently.
Still outstanding:
- Handling of 422 error responses from Vehicle Compliance Checker (in the case that the vehicle type is unknown).
- General error handling (e.g. the account ID is not identified, no vehicles have been registered for the account).


src/main/java/uk/gov/caz/psr/controller/AccountControllerApiSpec.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  @ApiOperation(
+      value = "${swagger.operations.accounts.vehicle-retrieval.description}",
+      response = VehicleRetrievalResponseDto.class
+  )
+  @ApiResponses({
+      @ApiResponse(code = 500, message = "Internal Server Error / No message available"),
+      @ApiResponse(code = 400, message = "Bad Request (the request is missing a mandatory "
+          + "element)"),
+      @ApiResponse(code = 401, message = "Unauthorized"),
```
plenza: I don't think this endpoint will ever return 401 because it is internal one which is unsecured

src/main/java/uk/gov/caz/psr/controller/AccountControllerApiSpec.java
```diff
@@ -0,0 +1,72 @@
...
+      @ApiImplicitParam(name = "pageSize",
+          required = true,
+          value = "The size of the page to retrieve",
+          paramType = "query"),
+      @ApiImplicitParam(name = "zones",
+          required = true,
+           value = "The clean air zones for which to return charges",
+           paramType = "query")
+  })
+  @GetMapping("/{accountId}/charges")
```
plenza: Is there any reason this endpoint is called /charges? I think what we return here is list of fleet vehicles with compliance results in each CAZ?

src/main/java/uk/gov/caz/psr/controller/AccountControllerApiSpec.java
```diff
@@ -0,0 +1,72 @@
...
+  ResponseEntity<VehicleRetrievalResponseDto> retrieveVehiclesAndCharges(
+      @PathVariable("accountId") UUID accountId, 
+      @RequestParam(required = true) Map<String, String> queryStrings) throws IOException;
```
plenza: Each parameter (pageNumber, pageSize and zones) can be separately added to this method signature. It will simplify implementation in `AccountController`

src/main/java/uk/gov/caz/psr/controller/AccountsController.java
```diff
@@ -0,0 +1,72 @@
...
+    validateRequest(queryStrings);
+    
+    AccountVehicleRetrievalResponse accountVehicleRetrievalResponse = 
+        accountService.retrieveAccountVehicles(accountId, queryStrings.get("pageNumber"),
+            queryStrings.get("pageSize"));
+    
+    List<ComplianceResultsDto> results = vehicleComplianceRetrievalService
+        .retrieveVehicleCompliance(
+        accountVehicleRetrievalResponse.getVrns(), 
+        queryStrings.get("zones"));
```
plenza: I may be wrong but isn't frontend expecting results for all known CAZes? So it was Payments-API responsibility to query VCCS for CAZes and then calculate compliances?

src/main/java/uk/gov/caz/psr/controller/AccountsController.java
```diff
@@ -0,0 +1,72 @@
...
+public class AccountsController implements AccountControllerApiSpec {
+
+  public static final String ACCOUNTS_PATH = "/v1/accounts";
+  
+  private final VehicleComplianceRetrievalService vehicleComplianceRetrievalService;
+  private final AccountService accountService;
+
+  @Override
+  public ResponseEntity<VehicleRetrievalResponseDto> retrieveVehiclesAndCharges(
+      UUID accountId, Map<String, String> queryStrings) throws IOException {
```
plenza: Why are you throwing IOException from this method?

src/main/java/uk/gov/caz/psr/dto/VehicleRetrievalResponseDto.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.caz.psr.dto;
+
+import java.util.List;
+import lombok.Builder;
+import lombok.Value;
+import uk.gov.caz.definitions.dto.ComplianceResultsDto;
+
+@Builder
+@Value
+public class VehicleRetrievalResponseDto {
```
plenza: Missing Javadocs

src/main/java/uk/gov/caz/psr/repository/AccountsRepository.java
```diff
@@ -0,0 +1,25 @@
...
+   *
+   * @return {@link Call}
+   */
+  @Headers("Accept: application/json")
+  @GET("v1/accounts/{accountId}/vehicles")
+  Call<AccountVehicleRetrievalResponse> getAccountVehicleVrns(
+      @Path("accountId") UUID accountId,
+      @Query("pageNumber") String pageNumber,
+      @Query("pageSize") String pageSize
+    );
```
plenza: Please add method to provide sync response and use it in `AccountService` - similar to this [idiom](https://github.com/InformedSolutions/JAQU-CAZ-Java-Common/blob/a5219d5cdb9b50304e284e8a4e5f02f75796e0f6/rest-async/src/test/java/uk/gov/caz/async/rest/AsyncClientDemoTest.java#L37)

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -0,0 +1,22 @@
...
+import java.io.IOException;
+import java.util.UUID;
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Service;
+import uk.gov.caz.psr.dto.AccountVehicleRetrievalResponse;
+import uk.gov.caz.psr.repository.AccountsRepository;
+
+@AllArgsConstructor
+@Service
+public class AccountService {
```
plenza: Missing Javadocs

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -0,0 +1,22 @@
...
+  
+  private final AccountsRepository accountsRepository;
+
+  public AccountVehicleRetrievalResponse retrieveAccountVehicles(UUID accountId,
+      String pageNumber, String pageSize) throws IOException {
+    return accountsRepository.getAccountVehicleVrns(accountId, pageNumber, pageSize).execute()
+        .body();
+  }
+
```
plenza: Please drop this line

src/main/java/uk/gov/caz/psr/repository/VccsRepository.java
```diff
@@ -33,4 +44,16 @@
...
+  /**
+   * Wraps REST API call in {@link AsyncOp} making it asynchronous.
+   *
+   * @param correlationId for correlation
+   * @param vrn the vrn to find compliance of
+   * @return {@link AsyncOp} with prepared REST call.
+   */
+  default AsyncOp<ComplianceResultsDto> findComplianceAsync(String vrn, String zones,
+      String correlationId) {
+    return AsyncOp.from("VCCS: " + correlationId, findCompliance(vrn, zones, correlationId));
```
plenza: Something is not right here: correlationId should match the one that initiated the request (from the frontend). So all (pageSize) async ops to VCCS will have the same key "VCCS: sameCorrelationId". It may be better to calc some hashCode from vrn (String) maybe?

src/main/resources/application.yml
```diff
@@ -6,6 +6,9 @@ springfox:
     swagger:
       v2:
         path: /v1/swagger-docs
+        
+server:
```
plenza: This should not be here. If anyone wants to use different port he/she should use env var before starting the application

src/main/resources/application.yml
```diff
@@ -29,7 +32,9 @@ services:
     rootUrl: https://publicapi.payments.service.gov.uk
     api-key: to-be-replaced-by-external-value
   vehicle-compliance-checker:
-    root-url: http://localhost:1090
+    root-url: http://localhost:8080
+  accounts:
+    root-url: http://localhost:8082
```
plenza: This again is fixed to your localhost setup which may not fit into other setup

kingarj: Good spot - is it best to leave these blank then?

src/main/java/uk/gov/caz/psr/repository/VccsRepository.java
kingarj: This should be resolved with the fix you suggested for handling 422s (creating a hash of UUIDs mapping to VRNs). I am working on a new branch at the moment to tackle this and other error responses.

src/main/java/uk/gov/caz/psr/service/exception/ExternalServiceCallException.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.caz.psr.service.exception;
+
+import java.util.Optional;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value = HttpStatus.SERVICE_UNAVAILABLE, reason = "Service unavailable")
+public class ExternalServiceCallException extends RuntimeException {
```
plenza: Please make it extend `ApplicationRuntimeException` just like [here](https://github.com/InformedSolutions/JAQU-CAZ-Payments-API/blob/e0b789878533f4d8f9d5ca583f28390823255718/src/main/java/uk/gov/caz/psr/service/exception/PaymentNotProcessedException.java#L13)

src/main/java/uk/gov/caz/psr/service/VehicleComplianceRetrievalService.java
```diff
@@ -0,0 +1,84 @@
...
+   * Coordinate asynchronous requests to the vehicle checker to retrieve compliance
+   * information on a list of VRNs.
+   * @param vrns a list of vrns
+   * @param zones a list of zones to check compliance for
+   * @return a list of compliance results sorted by vrn
+   */
+  public List<ComplianceResultsDto> retrieveVehicleCompliance(List<String> vrns, String zones) {
+    List<AsyncOp<ComplianceResultsDto>> complianceResultResponses = vrns.stream()
+        .map(vrn -> vccsRepository
+            .findComplianceAsync(vrn, zones, MdcCorrelationIdInjector.getCurrentValue()))
```
plenza: CorrelationId should be taken automatically in injector in `AsyncRestConfiguration` so can be removed completely from here.

src/main/java/uk/gov/caz/psr/controller/AccountControllerApiSpec.java
cruddasj: @plenza - the endpoint exposes both chargeable and non-chargeable vehicles for the account. In the event a non-compliant vehicle is included, this returns the numeric charge value so /charges makes sense 👍

src/main/resources/application.yml
plenza: @kingarj - You can put values you use during integration testing. So 1091 here. It will be invalid when no MockServer is running so it will force users to override by env var or setup ITs env before doing manual runs.

src/main/java/uk/gov/caz/psr/controller/AccountControllerApiSpec.java
kingarj: @plenza I did have it that way originally, however it seems to be much more straightforward to accept a query string map in terms of returning bespoke 400 error messages (otherwise Spring's request bindings returns a 400 error code but without a message).

PR #167, opened 2020-02-17
CAZ-1997 Allow payment initiation to know user who does the payment -…

radekmorytko:
… add `userId` attribute to init payment request

### DIFF
--- a/src/main/java/uk/gov/caz/psr/model/Payment.java
+++ b/src/main/java/uk/gov/caz/psr/model/Payment.java
@@ -24,7 +24,7 @@ public class Payment {
    * The unique payment identifier from GOV UK Pay service.
    */
   String externalId;
-  
+
   /**
    * The central reference number of the payment.
    */
@@ -85,6 +85,12 @@ public class Payment {
   UUID cleanAirZoneId;
 
   /**
+   * An identifier of the account/user which is making the payment. {@code null} for citizen
+   * journeys.
+   */
+  UUID userId;
+
+  /**
    * An overridden lombok's builder.
    */
   public static class PaymentBuilder {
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/caz/psr/model/Payment.java
```diff
@@ -84,6 +84,11 @@
    */
   UUID cleanAirZoneId;
 
+  /**
+   * An identifier of the account which is making the payment. {@code null} for citizen journeys.
+   */
+  UUID userId;
```
plenza: Maybe we want here to have an Optional? Or just the getter to be Optional? However I'm not sure about that, Optionals should really be used for method returns not as fields.

src/main/java/uk/gov/caz/psr/service/InitiatePaymentService.java
```diff
@@ -55,6 +57,9 @@ private Payment buildPayment(InitiatePaymentRequest request) {
         .totalPaid(calculateTotal(request.getTransactions()))
         .entrantPayments(Collections.emptyList())
         .cleanAirZoneId(request.getCleanAirZoneId())
+        .userId(StringUtils.hasText(request.getUserId())
```
plenza: Because DTO gets into the services layer, you need to do this silly conversion here. It should already be converted to model object.

src/main/java/uk/gov/caz/psr/model/Payment.java
radekmorytko: Ideally Lombok should take care of that and create a getter that returns `Optional` for a nullable field. However, this feature is not implemented: https://github.com/rzwitserloot/lombok/issues/1957
I'm not sure as well. For now I'd leave as it is to be consistent with the other fields.

src/main/java/uk/gov/caz/psr/service/InitiatePaymentService.java
radekmorytko: I created a dedicated utility class that deals with conversion to `Payment`. Additionally, there was a need to create a separate model for a single `transaction` (i.e. entrant payment)

PR #175, opened 2020-02-25
Determine page of chargeable vehicles

kingarj:
This PR comprises:
-  happy path functionality for fetching only chargeable vehicles for inclusion in the payments matrix.
- checking for entrant payments that have already been made for that particular combination of vrn, date (in the 13 day payment window) and Clean Air Zone.


src/it/java/uk/gov/caz/psr/ExternalCallsIT.java
```diff
@@ -64,6 +75,44 @@ public void mockVccsComplianceCallError(String vrn, int statusCode) {
...
+          parameter), exactly(1))
+      .respond(response(responseFile, 200));
+  }
+  
+  public void mockAccountServiceChargeableVehiclesCallWithoutCursor(String accountId, 
+      String responseFile) {
+    accountsMockServer
+      .when(requestGet("/v1/accounts/" + accountId + "/vehicles/sorted-page"), exactly(1))
+      .respond(response(responseFile, 200));
+  }
```
plenza: I'm wondering if word "Charges" and "Chargeable" has any meaning in Accounts context. Accounts returns list (page) of Fleet vehicles and that's all they do. The "chargeability" aspect is added later by Payment service logic. So maybe we should name these methods more clearly?

src/it/java/uk/gov/caz/psr/RetrieveChargeableAccountVehiclesIT.java
```diff
@@ -0,0 +1,217 @@
...
+      .forVrn("LMN234")
+      .forCleanAirZoneId(CLEAN_AIR_ZONE_ID)
+      .whenRequestIsMadeToRetrieveChargeableAccountVehicles()
+      .then()
+      .responseIsReturnedWithHttpOkStatusCode()
+      .responseContainsExpectedData(chargeable, null, chargeable.get(chargeable.size() - 1));
+  }
+  
+  @Test
+  public void shouldReturnHttpOkWhenEntrantPaymentsExist() throws JsonProcessingException {
```
plenza: Just to confirm: for now this is only a placeholder test because it will extended with paid days on a later stage?

src/it/java/uk/gov/caz/psr/RetrieveChargeableAccountVehiclesIT.java
```diff
@@ -0,0 +1,217 @@
...
+  @Test
+  public void shouldReturnHttpOkAndNoFirstVrnWhenNoFurtherPreviousVrnsFound() 
+      throws JsonProcessingException {
+    mockAccountServiceChargeableVehiclesCall(
+        ACCOUNT_ID, "LMN234", PREVIOUS_CURSOR_RESPONSE);
+    mockAccountServiceChargeableVehiclesCall(
+        ACCOUNT_ID, "ABC123", EMPTY_CURSOR_RESPONSE);
+    List<String> vrns = getAccountVehicleVrnsFromFile(PREVIOUS_CURSOR_RESPONSE);
+    List<String> chargeable = new ArrayList<String>();
+    for (int i = 0; i < vrns.size(); i++) {
```
plenza: There is some duplication with previous test: it would be good to keep them as neat as possible so maybe you could extract common setup to private helpers?

src/it/java/uk/gov/caz/psr/journeys/RetrieveChargeableAccountVehiclesJourneyAssertion.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public RetrieveChargeableAccountVehiclesJourneyAssertion forCleanAirZoneId(String cleanAirZoneId) {
+    this.cleanAirZoneId = cleanAirZoneId;
+    return this;
+  }
+  
+  public RetrieveChargeableAccountVehiclesJourneyAssertion 
+  whenRequestIsMadeToRetrieveChargeableAccountVehicles() {
+    RestAssured.basePath = AccountsController.ACCOUNTS_PATH;
```
plenza: This should probably be placed in the test setup to be more visible

src/it/java/uk/gov/caz/psr/RetrieveChargeableAccountVehiclesIT.java
```diff
@@ -0,0 +1,217 @@
...
+    givenAssertion()
+      .forAccountId(ACCOUNT_ID)
+      .forDirection("next")
+      .forPageSize("10")
+      .forVrn("ABC123")
+      .forCleanAirZoneId(CLEAN_AIR_ZONE_ID)
+      .whenRequestIsMadeToRetrieveChargeableAccountVehicles()
+      .then()
+      .responseIsReturnedWithHttpOkStatusCode()
+      .responseContainsExpectedData(chargeable, vrns.get(0), null);
```
plenza: last vrn is `null`? Is this correct?

src/it/java/uk/gov/caz/psr/journeys/RetrieveChargeableAccountVehiclesJourneyAssertion.java
```diff
@@ -0,0 +1,103 @@
...
+  public RetrieveChargeableAccountVehiclesJourneyAssertion responseIsReturnedWithHttpOkStatusCode() {
+    this.responseDto = response.statusCode(HttpStatus.OK.value())
+        .header(Constants.X_CORRELATION_ID_HEADER, CORRELATION_ID).extract()
+        .as(ChargeableAccountVehicleResponse.class);
+    return this;
+  }
+
+  public void responseContainsExpectedData(List<String> expectedVrns, 
+      String firstVrn, String lastVrn) {
+    List<PaidPaymentsResult> results = this.responseDto.getPaidPayments().getResults();
```
plenza: I don't understand we `getPaidPayments` are being queried here. I thought it is supposed to list all vehicles that need to be paid for not already paid ones?

src/main/java/uk/gov/caz/psr/dto/PaidPaymentsResponse.java
```diff
@@ -31,6 +31,7 @@ public static PaidPaymentsResponse from(Map<String, List<EntrantPayment>> result
     List<PaidPaymentsResult> mappedResults = results.entrySet()
         .stream()
         .map(entry -> buildPaidPaymentResultFrom(entry.getKey(), entry.getValue()))
+        .sorted((object1, object2) -> object1.getVrn().compareTo(object2.getVrn()))
```
plenza: You may want to try `.sorted(Comparator.comparing(PaidPaymentsResponse::getVrn)` for some more readibility

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -50,6 +59,95 @@ public AccountVehicleRetrievalResponse retrieveAccountVehicles(UUID accountId,
...
+   */
+  public PaidPaymentsResponse retrieveChargeableAccountVehicles(UUID accountId, String direction, 
+      int pageSize, String vrn, String cleanAirZoneId) {
+    List<String> results = new ArrayList<String>();
+    Boolean lastPage = false;
+    // initialise cursor at first VRN
+    String vrnCursor = vrn;
+    
+    while (results.size() < (pageSize + 1) && !lastPage) {
+      // get triple the number of vrns as will be on page to reduce overall request numbers
```
plenza: Good idea!

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -50,6 +59,95 @@ public AccountVehicleRetrievalResponse retrieveAccountVehicles(UUID accountId,
...
+      List<String> accountVrns = getAccountVrns(accountId, direction, pageSize * 3, vrnCursor);
+      
+      // check if the end of pages has been reached, if not set new cursor
+      if (accountVrns.size() < pageSize) {
```
plenza: Should this perhaps be `pageSize * 3`? Not a bug but a small optimization which may cut last 2 calls.

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -50,6 +59,95 @@ public AccountVehicleRetrievalResponse retrieveAccountVehicles(UUID accountId,
...
+        lastPage = true;
+      } else {
+        vrnCursor = getVrnCursor(accountVrns, direction);
+      }      
+      
+      List<String> chargeableVrns = getChargeableVrnsFromVcc(accountVrns, cleanAirZoneId);
+      
+      if (! chargeableVrns.isEmpty()) {
```
plenza: Please remove space after !

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -50,6 +59,95 @@ public AccountVehicleRetrievalResponse retrieveAccountVehicles(UUID accountId,
...
+    Response<List<String>> accountsResponse = accountsRepository
+        .getAccountVehicleVrnsByCursorSync(accountId, direction, pageSize, vrnCursor);
+    return accountsResponse.body();
+  }
+
+  private String getVrnCursor(List<String> accountVrns, String direction) {
+    Collections.sort(accountVrns);
+    if (direction.equals("next")) {
+      return accountVrns.get(accountVrns.size() - 1);
+    } else if (direction.equals("previous")) {
```
plenza: You can remove else keywords due to returns inside ifs

cruddasj:
@kingarj - as discussed yesterday, in the event we can't determine the vehicle type (r.g. 404 or 422), the vehicle should not appear in the list from the chargeable vehicles endpoint. Please can you revise the logic and action @plenza's review comments today?

kingarj:
@plenza many thanks for your review, please hold fire on re-review as I am just looking at a small improvement to batch asynchronous requests to VCCS.

src/it/java/uk/gov/caz/psr/journeys/RetrieveChargeableAccountVehiclesJourneyAssertion.java
kingarj: That is poor naming convention on my part - @cruddasj is renaming this to chargeableAccountVehicles in a forthcoming PR.

src/it/java/uk/gov/caz/psr/RetrieveChargeableAccountVehiclesIT.java
kingarj: I wasn't calling the right assertion method here - this now checks that the entrant payment (as inserted into the entrant_payment table at the beginning of the test, has been correctly returned and associated with the correct VRN.

kingarj:
@plenza I've addressed your comments, and added some small enhancements such as the piece around excluding unknown and undeterminable types of vehicles. Are you happy to give this a quick re-review?

src/main/java/uk/gov/caz/psr/service/AccountService.java
```diff
@@ -50,6 +59,118 @@ public AccountVehicleRetrievalResponse retrieveAccountVehicles(UUID accountId,
...
+        LocalDate.now().minusDays(6),
+        LocalDate.now().plusDays(6),
+        UUID.fromString(cleanAirZoneId));
+  }
+
+  private List<String> getChargeableVrnsFromVcc(List<String> accountVrns, 
+      String cleanAirZoneId, int pageSize) {
+    List<String> results = new ArrayList<String>();
+    // split accountVrns into chunks
+    List<List<String>> accountVrnChunks = Lists.partition(accountVrns, 10);
```
cruddasj: Could we pull this currently fixed value from the querystring param received at the controller layer? That way if we needed to change the size of how many vehicles are presented, only a UI layer update would be needed. Not sure if this is the same as the pageSize variable on this method signature?

kingarj: Hmm, my concern with that would be that if you bumped the pageSize var up to say 100 (for instance, if the test team was doing some robust testing of this endpoint), you could end up making that number of asynchronous batch requests to VCCS. What about a variable pulled from application.yml?

cruddasj: I think the challenge we would have is that in the above scenario we would face a problem that the application wouldn't be consistent through between front-end and back-end...

cruddasj: I think the better approach is to pull it from the variable and we make sure the UI can't be manipulated by query string modification

PR #182, opened 2020-03-06
Add proxy endpoint for retrieving unknown vehicle compliance

cruddasj:
Add proxy endpoint for retrieving unknown vehicle compliance


cruddasj:
Java is now done - just IAC to go!

PR #183, opened 2020-03-09
Feature/caz 2121/tariff proxy caching

cruddasj:
Add caching for tariffs alongside cache invalidation controller.


src/main/java/uk/gov/caz/psr/service/CleanAirZoneService.java
```diff
@@ -27,28 +31,34 @@
...
-  public Response<CleanAirZonesResponse> fetchAll() {
+  @Cacheable(value = "cleanAirZones")
+  public CacheableResponseDto<CleanAirZonesResponse> fetchAll()
+      throws JsonProcessingException {
     try {
       log.debug("Fetching all clean air zones from VCCS");
-      return vccsRepository.findCleanAirZonesSync();
+      Response<CleanAirZonesResponse> response =
+          vccsRepository.findCleanAirZonesSync();
+      return CacheableResponseDto.<CleanAirZonesResponse>builder()
```
cruddasj: Note this has been the only way identified to cache a retrofit response (by effectively wrapping it) given the Response class is final and non-serializable. The CacheableResponseDto (placed in common lib) accepts a type argument, which can then be used higher up in the stack (e.g. controller layer) for returning as a typed ResponseEntity.

PR #184, opened 2020-03-09
Add logging to vccs repository

cruddasj:
Add logging to VCCS repository to support diagnostics and test evidence (by dynamically resolving slf4j DI logger implementation).


cruddasj:
@plenza - tagging in as an FYI on logging with interface declarations 👍

PR #187, opened 2020-03-11
Permit payments to proceed in the event of one being in-flight""

cruddasj:
Allow new payments to be initiated when existing payments are in-flight


cruddasj:
Re-introducing as this implicates Fleets and the payment matrix which amplifies the issue at hand.

PR #190, opened 2020-03-12
Use bulk compliance checker endpoint instead of async requests

kingarj:


cruddasj:
Good to merge into dev 👍

PR #198, opened 2020-03-19
Adjust lambda initialisation method

cruddasj:
Having observed cold starts in dev, this change mirrors those introduced to VCCS under InformedSolutions/JAQU-CAZ-Vehicle-Compliance-Checker-API#440.


cruddasj:
@plenza, @phan-anh-tuan - FYI on the above.

PR #200, opened 2020-03-25
CAZ-2177 Create a new endpoint for getting clean air zones alongside …

radekmorytko:
…direct debit mandates


src/main/resources/application.yml
```diff
@@ -14,7 +14,7 @@ spring:
     enabled: true
   datasource:
     driver-class-name: org.postgresql.Driver
-    url: jdbc:postgresql://localhost:5433/vehicle_compliance
+    url: jdbc:postgresql://localhost:5432/vehicle_compliance
```
dratka: port should not be updated in this commit.

radekmorytko: Agree, changed :+1:

PR #201, opened 2020-03-27
[CAZ-2171] Create DirectDebitMandate endpoint

dratka:
Jira Card: https://eaflood.atlassian.net/browse/CAZ-2171


src/main/java/uk/gov/caz/psr/controller/DirectDebitMandatesControllerApiSpec.java
```diff
@@ -42,4 +47,28 @@
...
+      @ApiResponse(code = 400, message = "Missing Correlation Id header or invalid params")})
+  @ApiImplicitParams({@ApiImplicitParam(name = "X-Correlation-ID",
+      required = true,
+      value = "UUID formatted string to track the request through the enquiries stack",
+      paramType = "header")})
+  @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE)
+  @ResponseStatus(HttpStatus.CREATED)
+  ResponseEntity<CreateDirectDebitMandateResponse> createDirectDebitMandate(
+      @PathVariable("accountId") UUID accountId,
+      @Valid @RequestBody CreateDirectDebitMandateRequest request);
```
radekmorytko: If we have our custom validators I think `@Valid` annotation is not necessary.

src/main/java/uk/gov/caz/psr/repository/AccountsRepository.java
```diff
@@ -123,4 +127,29 @@
       throw new ExternalServiceCallException(e.getMessage());
     }
   }
+
+  /**
+   * Method to create DirectDebitMandate for account.
+   *
+   * @return {@link Call}
+   */
+  @Headers("Accept: application/json")
```
radekmorytko: As this is a `POST` request don't we need the `Content-Type` header?

src/main/java/uk/gov/caz/psr/service/directdebit/DirectDebitMandatesService.java
```diff
@@ -51,6 +52,21 @@
     }
   }
 
+  /**
+   * Creates DirectDebitMandate in External Payment provider, store details in AccountsAPI and
+   * returns nextUrl to confirm Mandate creation on the frontend.
+   */
+  public String createDirectDebitMandate(UUID cleanAirZoneId, UUID accountId, String returnUrl) {
```
radekmorytko: We do not have any logging here - maybe we can add something just to have a trace of the operation?

PR #203, opened 2020-03-30
CAZ-2253 - Return PaymentMethod from Vehicle Entrants endpoint

plenza:
CAZ-2253 - Return PaymentMethod from Vehicle Entrants endpoint

**Business logic changes:**

We need to have "String: paymentMethod" in response from POST /vehicle-entrants endpoint. There are some rules:

- It should be "null" if this is first vehicle entrant and it has not been paid in advance, so we don't have any EntrantPayment entities yet
- It should be "null" if EntrantPayment exists but is not "PAID"
- It should be "null" if EntrantPayment exists, is "PAID" but there is no corresponding latest Payment record from which we fetch PaymentMethod. Not sure if this case is even possible, but it has been covered and tested.
- It should be "card" if EntrantPayment exists and has been paid by credit or debit card - so it is in "PAID" status and we have latest Payment with proper PaymentMethod
- It should be "direct_debit" if EntrantPayment exists and has been paid by direct debit - so it is in "PAID" status and we have latest Payment with proper PaymentMethod


src/main/java/uk/gov/caz/psr/service/EntrantPaymentService.java
```diff
@@ -75,4 +123,70 @@ private EntrantPayment buildNewEntrantPayment(VehicleEntrantDto vehicleEntrantDt
...
+  }
+
+  /**
+   * If {@link EntrantPayment} is present and is "PAID" we try to fetch corresponding matching
+   * latest {@link Payment} record. If it is present we can fetch PaymentMethod to return. If it is
+   * not present we will return "null" as PaymentMethod.
+   */
+  private void processWithPotentialPayment(List<EntrantPaymentWithLatestPaymentDetailsDto> result,
+      EntrantPayment entrantPayment) {
+    Optional<Payment> payment = findLatestPaymentIfPresent(entrantPayment);
```
plenza: Note for reviewers: This implementation is suboptimal. First there is query for `EntrantPayment` and then for each `PAID` another query for `Payment` with join over `Match` table. So we have N^2 queries. This will suck, but according to CAZ-2253 this is OK for this initial implementation.

src/test/java/uk/gov/caz/psr/service/EntrantPaymentServiceTest.java
```diff
@@ -160,4 +286,28 @@ private EntrantPayment buildCazEntrantPaymentToInsert() {
         .updateActor(EntrantPaymentUpdateActor.VCCS_API)
         .build();
   }
+
+  private void mockNoMatchingPayment() {
+    when(paymentRepository.findByEntrantPayment(ENTRANT_PAYMENT_ID)).thenReturn(Optional.empty());
+  }
+
+  private void mockMatchingPaymentWithPaymentMethod(
+      uk.gov.caz.psr.model.PaymentMethod paymentMethod) {
```
jaroslawpiotrowski: Some import problem

plenza: Indeed. Will be fixed

plenza: Done

PR #210, opened 2020-04-01
CAZ-2251 Payments API - update GET payment-status to return payment m…

adamPM1:
…ethod and mandate ID

### DIFF
--- a/src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
+++ b/src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
@@ -1,10 +1,14 @@
 package uk.gov.caz.psr.dto;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import io.swagger.annotations.ApiModelProperty;
+import javax.annotation.Nullable;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.NotNull;
 import lombok.Builder;
 import lombok.Value;
+import uk.gov.caz.psr.model.PaymentMethod;
 import uk.gov.caz.psr.model.PaymentStatus;
 
 @Value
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
```diff
@@ -41,6 +50,12 @@ public static PaymentStatusResponse from(PaymentStatus paymentStatus) {
         .paymentProviderId(paymentStatus.getExternalId())
         .caseReference(paymentStatus.getCaseReference())
         .paymentMethod(ChargeSettlementPaymentMethod.from(paymentStatus.getPaymentMethod()))
+        .paymentMandateId(
```
radekmorytko, line 53: If this value is `null` then the key `paymentMandateId` is still rendered in the response, I think we should annotate the class with `@JsonInclude(Include.NON_NULL)` to include only the non-null values. I assume this is the requirement:
> add optional paymentMandateId to response (if payment has been made by direct debit)

src/main/resources/application.yml
```diff
@@ -72,6 +72,8 @@ swagger:
         vrn: Vehicle registration number
         date-of-caz-entry: ISO 8601 formatted date string indicating the date that the vehicle identified by the VRN attribute entered the CAZ.
         payment-method: Method of the payment
+        payment-mandate-id: An mandate ID
```
radekmorytko, line 75: Maybe either remove _An_ or change to _Direct debit mandate identifier_?

adamPM1: changed, thanks :)

src/main/java/uk/gov/caz/psr/dto/PaymentStatusResponse.java
adamPM1: I put suggested annotation on the top of field since I am not sure if NON_NULL should be applied to all fields.

PR #212, opened 2020-04-02
CAZ-2252 extend payment info response

jarnie86:


src/main/java/uk/gov/caz/psr/dto/PaymentInfoResponse.java
```diff
@@ -44,6 +47,14 @@
     @ApiModelProperty(value = "${swagger.model.descriptions.payment-info.payment-method}")
     ChargeSettlementPaymentMethod paymentMethod;
 
+    @Nullable
+    @JsonInclude(Include.NON_NULL)
+    @ApiModelProperty(value = "${swagger.model.descriptions.payment-info.payment_mandate_id}")
```
radekmorytko, line 52: I cannot see those keys have been added to `application.yml`. Additionally, I would change the key to `payment-mandate-id` to follow the current convention.

PR #229, opened 2020-04-29
Expose dd creation date

cruddasj:
This change is to support a JIRA task which sees revisions to content (where the created date gets shown). I've considered whether this should be added under the accounts API, however, given we are refreshing status anyway by calling out to GOV.UK PAY when hitting the /payments/accounts/<<account-id>>/direct-debit-mandates/<<mandate-id>> endpoint calls, it seems pulling the creation date through at the same time should not make any difference in performance.

The response for this endpoint is fine (and includes an iso stamped created value) as expected.

The below response (on the mandates across different CAZ's), however, does seem odd (using the mandates fetched from the account service). 

With this in mind, an approach that does store things in the Account service may still be pursued tomorrow but PR is being raised for team visibility. This latter approach does also have some downsides (as the mandate can only be created after a success call from GOV.UK PAY is returned, allowing the created date to be set properly).

` "cleanAirZones": [
        {
            "cazId": "5cd7441d-766f-48ff-b8ad-1809586fea37",
            "cazName": "Birmingham",
            "mandates": [
                {
                    "id": "8cie6doealplvd11at1hpjs6vo",
                    "status": "PENDING",
                    "created": null
                }
            ]
        },
        {`

### DIFF
--- a/src/it/resources/data/json/response/direct-debit-mandates-for-caz-response.json
+++ b/src/it/resources/data/json/response/direct-debit-mandates-for-caz-response.json
@@ -2,15 +2,18 @@
   "mandates": [
     {
       "id": "i4nuo03jginfke5c3ebrvgig6a",
-      "status": "CREATED"
+      "status": "CREATED",
+      "created": "2020-03-19"
     },
     {
       "id": "k4nuo03jginfke5c3ebrvgig6a",
-      "status": "CREATED"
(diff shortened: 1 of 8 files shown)


cruddasj:
On reflection, keeping the created date in the account service would be more useful for support purposes so I'm closing this PR. Please ignore....

PR #282, opened 2020-07-20
CAZB-2317 Timestamp inconsistency in VEI and CSI requests and header

jaroslawpiotrowski:


src/main/java/uk/gov/caz/psr/controller/ChargeSettlementControllerApiSpec.java
```diff
@@ -75,7 +75,7 @@
   @GetMapping(ChargeSettlementController.PAYMENT_INFO_PATH)
   ResponseEntity<PaymentInfoResponse> getPaymentInfo(@Valid PaymentInfoRequest paymentInfoRequest,
       BindingResult bindingResult, @RequestHeader(Headers.X_API_KEY) UUID cleanAirZoneId,
-      @RequestHeader(Headers.TIMESTAMP) @DateTimeFormat(iso = DATE_TIME) LocalDateTime timestamp);
+      @RequestHeader("timestamp") @DateTimeFormat(iso = DATE_TIME) LocalDateTime timestamp);
```
radekmorytko, line 78: BTW It would be good to raise a separate PR to update the value of this constant

PR #286, opened 2020-07-20
[CAZB-2164] AccountsPayments endpoint

musztardem:
**JIRA Card**
https://eaflood.atlassian.net/browse/CAZB-2164

----

**Changes description**
* The following changes exposes endpoint which returns data fetched from VCCS API (clean air zones data), Accounts API (users data) and from the database.
* The whole logic is stored in the `RetrieveSuccessfulPaymentsService` which exposes 2 public methods:
    * `retrieveForAccount` - for getting all users data based on provided `accountId`,
    * `retrieveForSingleUser` - for getting payments data for a single user associated with the `accountId`
* Data coming from the local database looks as follows:
    * Using `PaymentSummaryRepository#getTotalPaymentsCountForUserIds` the pagination data are fetched.
    * Using `PaymentSummaryRepository#getPaginatedPaymentSummaryForUserIds` the following payments data are fetched:
    ```
    {
        "paymentId": "XXXXXX",
        "entriesCount": 10,
        "totalPaid": 3,
        "cleanAirZoneId": null,
        "payerId": "XXXXXX"
    }
    ```
    Those data are stored in the `PaymentSummary` object. Note that this response has no `cleanAirZoneId` assigned.
    * Using `PaymentToCleanAirZoneMappingRepository#getPaymentToCleanAirZoneMapping` we get the information about Clean Air Zone associated with each payment:
     ```
    {
        "cleanAirZoneId": "XXXXXXX",
        "paymentId": "YYYYYYYY"
    }
    ```
    Those results are stored in `PaymentToCleanAirZoneMapping`
* Using `PaymentToCleanAirZoneMapping` the previously created `PaymentSummary` objects are enriched with `cleanAirZoneId` information.
* As we do not store information about clean air zone names and user names, those data are fetched from the AccountsAPI and VCCS - `PaymentSummary` objects are mapped to `EnrichedPaymentSummary` objects with changed attributes - `cleanAirZoneId` is mapped to `cazName` and `payerId` is mapped to `payerName`.

----

**Screenshots**
Screenshot 2020-07-17 at 16 25 22

### DIFF
--- a/src/main/java/uk/gov/caz/psr/controller/AccountsController.java
+++ b/src/main/java/uk/gov/caz/psr/controller/AccountsController.java
@@ -8,6 +8,7 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import lombok.AllArgsConstructor;
+import org.springframework.data.util.Pair;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.RestController;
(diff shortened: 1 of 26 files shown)


src/main/java/uk/gov/caz/psr/repository/AccountsRepository.java
```diff
@@ -180,4 +181,25 @@
       throw new ExternalServiceCallException(e.getMessage());
     }
   }
+
+  /**
+   * Method to get all users associated with the provided account.
+   */
+  @Headers({
+      "Accept: application/json",
+      "Content-Type: application/json"
```
radekmorytko, line 190: I think we don't need `Content-Type` header in `GET` requests, what do you think?

src/main/java/uk/gov/caz/psr/controller/AccountsController.java
```diff
@@ -143,6 +149,34 @@ private String checkDirectionQueryString(String direction, String vrn) {
         .body(response);
   }
 
+  @Override
+  public ResponseEntity<SuccessfulPaymentsResponse> retrieveSuccessfulPayments(UUID accountId,
+      Map<String, String> queryStrings) {
+
+    queryStringValidator.validateRequest(queryStrings,
```
radekmorytko, line 156: Minor: what do you think about extracting this to a private helper method, e.g. `validateRetrieveSuccessfulPaymentsRequest`?

PR #319, opened 2020-08-06
CAZB-2559 add operator id to t_payment table

jaroslawpiotrowski:


src/main/resources/db/changelog/changes/0009-1.0-add-operation-id-to-payment-table.yaml
```diff
@@ -0,0 +1,23 @@
...
+      changes:
+        - addColumn:
+            schemaName: CAZ_PAYMENT
+            tableName: T_PAYMENT
+            columns:
+              - column:
+                  name: OPERATOR_ID
+                  type: uuid
+                  constraints:
+                    unique: true
```
radekmorytko, line 14: I think this column should not be unique - if the same operator makes a second payment we will get an error.

jaroslawpiotrowski: You are right, thanks!

PR #322, opened 2020-08-07
rename dependabot file

ISJDF:


.drone2.yml
```diff
@@ -1693,6 +1784,24 @@ steps:
       event:
         - push
 
+  # subscribe to operational monitoring alert in api prod
+  - name: subscribe to operational monitoring alert in api prod
+    image: alpine:3.11.6
+    commands:
+      - apk add --update python3 groff
+      - pip3 install awscli
+      - python3 ./ci-cd-resources/subscribe_to_alert_topics.py -e freeable-memory -e 4xx -e usage -e evictions -e csv -i prod -i payments ISSupport-CAZ@informed.com
```
cruddasj, line 1793: Can we please remove this from the commit? It's also worth noting this needs a profile supplied in keep with the example at 
https://github.com/InformedSolutions/JAQU-CAZ-Retrofit-API/blob/344aee5cd559b33649b02238539bb46d721770bf/.drone2.yml#L1524

cruddasj (changes requested):

.drone2.yml
ISJDF: Removed

cruddasj (approved):

PR #327, opened 2020-08-11
CAZB-2587 references history endpoint

jaroslawpiotrowski:


src/main/java/uk/gov/caz/psr/dto/ReferencesHistoryResponse.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @ApiModelProperty(value = "${swagger.model.descriptions.references-history"
+      + ".central-reference-number}")
+  Long paymentReference;
+
+  @ApiModelProperty(value = "${swagger.model.descriptions.references-history.payment-provider-id}")
+  String paymentProviderId;
+
+  @ApiModelProperty(value = "${swagger.model.descriptions.references-history.payment-timestamp}")
+  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS")
```
lukaszkuczynski, line 30: I think Spec does not say to include milliseconds

src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
```diff
@@ -245,6 +274,9 @@ private MapSqlParameterSource toSqlParametersForInsert(Payment payment) {
 
     static final String SELECT_BY_ID =
         "SELECT " + ALL_PAYMENT_ATTRIBUTES + "FROM caz_payment.t_payment " + "WHERE payment_id = ?";
+
```
lukaszkuczynski, line 277: What about using `PaymentInfo` and Spring Data repositories?

src/main/java/uk/gov/caz/psr/util/ReferencesHistoryConverter.java
```diff
@@ -0,0 +1,94 @@
...
+        .lineItems(toVehicleEntrantPaymentsDetails(payment.getEntrantPayments()))
+        .build();
+  }
+
+  private void checkIfCazIsUnique(Payment payment) {
+    boolean isUniqueCaz = payment.getEntrantPayments()
+        .stream()
+        .map(EntrantPayment::getCleanAirZoneId)
+        .distinct()
+        .limit(2)
```
lukaszkuczynski, line 59: why do we have `limit` here?

jaroslawpiotrowski: We just need to check if there is more than one

src/main/java/uk/gov/caz/psr/dto/ReferencesHistoryResponse.java
jaroslawpiotrowski: Good catch, thanks!

src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
jaroslawpiotrowski: I would keep it for consistency because im using also `findById` from `PaymentRepository`

PR #345, opened 2020-09-04
Non jira/subscribe to monitoring

ISJDF:


cruddasj (approved):

ISJDF:
The python script is incorrect and will fail to find the alarms to subscribe to

PR #394, opened 2020-10-27
[CAZB-3307]  Charge settlement API returning XML response

jenniferbrown94:


cruddasj:
@jenniferbrown94 - can we please follow this with a commit that provides commentary on why this is used?

cruddasj (dismissed):

cruddasj (approved):

PR #455, opened 2021-02-05
CAZB-3943 Trigger from audit table after insert or update of CAZ entrant payment enriched with Payment info

lukaszkuczynski:
https://eaflood.atlassian.net/browse/CAZB-3943

a file to test is enclosed as an attachment to this ticket


musztardem (approved):

src/main/resources/db/changelog/changes/0021-1.0-details-trigger-updated-to-insert-payment-information.yaml
```diff
@@ -0,0 +1,13 @@
+databaseChangeLog:
+  - changeSet:
+      id: 0021-1.0
```
musztardem, line 3: One minor thing: I would add something as `0021-1.0-iod` or `0021-1.0-paymets` as recently I noticed that we store in the database changesets from all the services which cause ID duplications and can possibly cause some problems with management

musztardem (approved):

PR #460, opened 2021-02-15
[CAZB-3914] Add Refund/Chargeback history to the paymentReference endpoint

dratka:
Jira card: https://eaflood.atlassian.net/browse/CAZB-3914

image

### DIFF
--- a/src/main/resources/db/changelog/changes/0023-1.0-add-index-for-reference-history-serach.yaml
+++ b/src/main/resources/db/changelog/changes/0023-1.0-add-index-for-reference-history-serach.yaml
@@ -1,0 +1,21 @@
+databaseChangeLog:
+  - changeSet:
+      id: 0023-1.0-iod
+      author: dawid.ratka
+      preConditions:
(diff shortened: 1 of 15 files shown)


src/main/resources/db/changelog/changes/0023-1.0-add-index-for-reference-history-serach.yaml
```diff
@@ -0,0 +1,16 @@
+databaseChangeLog:
+  - changeSet:
```
musztardem, line 2: As I often run into problems with migrations idempotency, I would add the following precondition:
```yml
- preconditions:
    - onFail: MARK_RAN
    - sqlCheck:
        expectedResult: 0
        sql: SELECT COUNT(*) from pgindexes where schemaname = 'caz_payment_audit' AND indexname = 'payment_id_update_actor_payment_status'; 
```

musztardem (approved):

musztardem (approved):

PR #461, opened 2021-02-17
[CAZB-3999] Return Refund/Chargeback flag in operators-history GET endpoint

dratka:
Jira Card: https://eaflood.atlassian.net/browse/CAZB-3999
image

### DIFF
--- a/src/main/java/uk/gov/caz/psr/dto/historicalinfo/PaymentsInfoByOperatorResponse.java
+++ b/src/main/java/uk/gov/caz/psr/dto/historicalinfo/PaymentsInfoByOperatorResponse.java
@@ -1,6 +1,8 @@
 package uk.gov.caz.psr.dto.historicalinfo;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModelProperty;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/caz/psr/dto/historicalinfo/PaymentsInfoByOperatorResponse.java
```diff
@@ -105,5 +105,11 @@
     @ApiModelProperty(value =
         "${swagger.model.descriptions.payments-info-by-operator-id-response.status}")
     String paymentProviderStatus;
+
+    @ApiModelProperty(value = "${swagger.model.payments-info-by-operator-id-response.isRefunded")
+    boolean isRefunded;
```
musztardem, line 110: There's a little problem with attributes which starts with `isSomething` as Spring removes `is` prefix during serialization - you can check a problem description in this discussion: https://stackoverflow.com/questions/32270422/jackson-renames-primitive-boolean-field-by-removing-is

As we have already implemented a similar solution in other places, I would suggest resolving it in a similar manner:
* Add to those two attributes the following annotation `@JsonProperty(value = "isRefunded")`
* Annotate the class with `@JsonIgnoreProperties({"refunded", "chargedback"})` as the attributes annotation results in serializing both `isRefunded` and `refunded` attributes.

musztardem (dismissed):

musztardem (approved):

PR #518, opened 2021-05-04
[CAZB-4691] Add SSL support for elasticache

cruddasj:
Add SSL support for elasticache


cruddasj:
Please note that at this point in time the SSL flag is not configured on the Lambda environment variables - hence connections will not be SSL enabled until this is done.

proctorj (approved):

likeuwill (approved):

PR #566, opened 2021-07-20
[CAZB-3828] Charge Settlement API - Add parameter to search by central payment reference number

proctorj:
https://eaflood.atlassian.net/browse/CAZB-3828


src/main/java/uk/gov/caz/psr/dto/validation/constraint/ValidParameterCombinationPresentValidator.java
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * Checks if the passed {@code input} contains a valid combination of request parameters.
+   *
+   * @param input An input request which will be validated.
+   * @param context Validator context (unused).
+   * @return true if {@code input} contains at least on valid set of params.
+   */
+  @Override
+  public boolean isValid(PaymentStatusRequest input, ConstraintValidatorContext context) {
+    return input == null || input.getCazPaymentReference() != null
```
musztardem, line 20: this condition allows us to send `vrn` and `cazPaymentReference` parameters without providing `dateOfCazEntry` which I think is not valid as originally we were supposed to always send both `vrn` and `dateOfCazEntry`. 
I thought that we should be able to send this combination of parameters:
* `vrn` + `dateOfCazEntry`
* `vrn` + `dateOfCazEntry` + `cazPaymentReference`
* `cazPaymentReference`

src/main/java/uk/gov/caz/psr/model/PaymentInfoRequestAttributes.java
```diff
@@ -11,6 +11,8 @@
 @Value
 public class PaymentInfoRequestAttributes {
   String externalPaymentId;
+  
+  String cazPaymentReference;
```
musztardem, line 15: `cazPaymentReference` is an integer attribute in the database so I think we should have it defined also as `Integer` or `Long` here. When we send a string consisting on letters and not numbers we will get 500 exception instead of 400

src/main/java/uk/gov/caz/psr/repository/PaymentStatusRepository.java
```diff
@@ -53,24 +66,46 @@
...
+          selectByCazPaymentReferenceAndCazId(),
+          preparedStatement -> {
+            preparedStatement.setObject(1, cazId);
+            preparedStatement.setLong(2, Long.parseLong(cazPaymentReference));
+          },
+          ROW_MAPPER
+      );
+    } else {
+      LocalDate date = dateOfCazEntry != null ? LocalDate.parse(dateOfCazEntry) : null;
+      return jdbcTemplate.query(
```
musztardem, line 98: Currently, in a situation when we send e.g. `cazPaymentReference ` and `dateOfCazEntry` we are going to fall into this condition and search for a `vrn` that is `null` and that's really misleading, so we should adjust the attributes validation in `ValidParameterCombinationPresentValidator`

src/test/java/uk/gov/caz/psr/dto/validation/constraint/ValidParameterCombinationPresentValidatorTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    // when
+    boolean result = validator.isValid(request, context);
+
+    // then
+    assertThat(result).isTrue();
+  }
+
+  @Test
+  public void shouldReturnFalseTrueWhenVrnIsPresent() {
```
musztardem, line 60: So `False` or `True`? xD

src/test/java/uk/gov/caz/psr/dto/validation/constraint/ValidParameterCombinationPresentValidatorTest.java
```diff
@@ -0,0 +1,86 @@
...
+        .dateOfCazEntry("2019-11-01")
+        .build();
+
+    // when
+    boolean result = validator.isValid(request, context);
+
+    // then
+    assertThat(result).isFalse();
+  }
+}
```
musztardem, line 86: additional specs for the parameters conjunctions would be useful here

src/main/java/uk/gov/caz/psr/model/PaymentInfoRequestAttributes.java
musztardem: Same thing would apply to `PaymentStatusRequest` and `PaymentInfoRequestV1`

musztardem (changes requested):
couple changes

musztardem (approved):

PR #598, opened 2021-09-13
[CAZB-5389] Reconcile payment status based on PaymentStateCode in ExternalProvider

dratka:
Based on the API reference https://docs.payments.service.gov.uk/api_reference/#errors-caused-by-payment-statuses we need to check the payment status based on Payment State code to check if it was cancelled

https://eaflood.atlassian.net/browse/CAZB-5389


likeuwill (approved):
perfect ;)

PR #673, opened 2021-12-23
[CAZB-5881] Add hotfix logic to prod

IS-lamberto:


ci-cd-resources/hotfix-check.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+NORMAL_BRANCH="staging"
+HOTFIX_BRANCH="hotfix-staging"
```
kingarj, line 3: This should be `uat/`

PR #681, opened 2022-01-19
Promote Payments v2.9.0 to Pre-Production

proctorj:
Promote Payments v2.9.0 to Pre-Production


src/main/resources/application.yml
```diff
@@ -2,6 +2,9 @@ application:
   title: Payment Microservice
   emails-to-skip: emailtoskip@informed.com
 
+server:
```
kingarj, line 5: Hmm, is there any possibility that this knackers us?
@proctorj @cruddasj

PR #745, opened 2022-04-20
[CAZB-6545] Store Payment Failure Code

IS-lamberto:


src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
```diff
@@ -503,6 +518,8 @@ public Payment mapRow(ResultSet resultSet, int i) throws SQLException {
           .paymentProviderMandateId(resultSet.getString(Columns.PAYMENT_MANDATE_PROVIDER_ID))
           .telephonePayment(resultSet.getBoolean(Columns.TELEPHONE_PAYMENT))
           .emailConfirmationSent(resultSet.getBoolean(Columns.EMAIL_CONFIRMATION_SENT))
+          .externalPaymentStateCode(
```
kingarj, line 521: Is this not equivalent to calling `resultSet.getString(Columns.PAYMENT_PROVIDER_ERROR_CODE)` here?

src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
```diff
@@ -50,6 +50,7 @@ public class PaymentRepository {
   private final SimpleJdbcInsert simpleJdbcInsert;
   private final EntrantPaymentRepository entrantPaymentRepository;
   private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+  private String errorCode = null;
```
kingarj, line 53: Do we need to declare this at class level?

kingarj (changes requested):
Few minor points.

I suspect that in the audit table we're storing the old and new records as `JSONB` but can we just check that the new column is being correctly included in those audits? Should also form part of the development test checks.

src/main/java/uk/gov/caz/psr/repository/PaymentRepository.java
IS-lamberto: image

Trying to declare the variable 'errorCode' within the scope of the 'update' method requires that the variable is final, so it had to be taken out of the scope to resolve the issue.

IS-lamberto:
@kingarj Regarding the audit tables, the new column is successfully added to the logged_actions table in the new_data.

PR #755, opened 2022-05-11
Update version label to v2.13.0

likeuwill:


src/main/resources/application.yml
```diff
@@ -97,7 +97,7 @@ swagger:
   api:
     title: Payments API
     description: Payments API v2.12.0
```
arctindev, line 99: description: Payments API v2.13.0

arctindev (changes requested):

arctindev (approved):

PR #762, opened 2022-05-17
Deploy to Prod - v2.12.0

IS-lamberto:


cruddasj (approved):
Approved, however, I do want to see changes in the next release to remove the test-payments endpoint in the production API Gateway resources. Protections are in place by way of CIDR scoping to restrict traffic to the same VPC - but I would prefer that we don't provision these resources if the environment is pre-production or higher (as invocation can delete all payments along with associate audit data for a quoted VRN).

kingarj:
> Approved, however, I do want to see changes in the next release to remove the test-payments endpoint in the production API Gateway resources. Protections are in place by way of CIDR scoping to restrict traffic to the same VPC - but I would prefer that we don't provision these resources if the environment is pre-production or higher (as invocation can delete all payments along with associate audit data for a quoted VRN).

@cruddasj see my comment [here](https://github.com/InformedSolutions/JAQU-CAZ-IAC/pull/2528#pullrequestreview-975178283) - the paths are being created but no resources created for them on environments higher than SIT.

cruddasj:
> > Approved, however, I do want to see changes in the next release to remove the test-payments endpoint in the production API Gateway resources. Protections are in place by way of CIDR scoping to restrict traffic to the same VPC - but I would prefer that we don't provision these resources if the environment is pre-production or higher (as invocation can delete all payments along with associate audit data for a quoted VRN).
> 
> @cruddasj see my comment [here](https://github.com/InformedSolutions/JAQU-CAZ-IAC/pull/2528#pullrequestreview-975178283) - the paths are being created but no resources created for them on environments higher than SIT.

Okay - fair enough 👍 For cleanliness - if we can avoid creating the resource that would be great but appreciate it makes depends_on values a bit more tricky in terraform...

kingarj (approved):

PR #791, opened 2022-08-03
[CAZB-7213] Use cacheable method to get CAZ list for direct debit mandates

wojw5:
https://informedsolutions.atlassian.net/browse/CAZB-7213


bprzemek80 (approved):

src/it/java/uk/gov/caz/psr/directdebit/DirectDebitJourneyTestIT.java
```diff
@@ -724,4 +729,12 @@ private void executeSqlFrom(String classPathFile) {
   private String toJsonString(Object request) {
     return objectMapper.writeValueAsString(request);
   }
+
+  private void clearCleanAirZonesCache() {
```
wojw5, line 733: @FYI @dratka @bprzemek80 
This is a pretty ugly way of making sure we will clear the cache at the beginning of the test and at the end. It is because, if I understand correctly we are storing different values in the cache and we are not clearing them. If there is another, better way in the system to handle that please let me know, otherwise it can be the subject to discussion.

 This article gives some visibility on a potential solutions. Although it could be automated better probably. https://dzone.com/articles/spring-cache-and-integration-testing

wojw5: It actually can be the reason why Integration tests are failing sometimes randomly

wojw5: one more question is whether we should cache unsuccessful responses

wojw5: I used an already existing method to evict cache data explicitly and it seems to be the best solution.

dratka (approved):

PR #835, opened 2022-10-14
Deploy to Prod - v2.16.1

IS-lamberto:


kingarj (approved):
@cruddasj FYI - the significant change in the terraform plan here is that the secret for the Gov.UK Pay API keys and the GoCardless API keys is being replaced (as it has been added to terraform for BCDR reasons). The secret values have been checked and updated against the live equivalent as of Thursday and I've also taken a local backup of the current value of the secret as a failsafe.

cruddasj (approved):
Approved based on TF plan and IAC commit inspection (paired with fallbacks) - however, @kingarj - we should discuss this tomorrow. As a priority - given API keys have now been deposited to source control in the IAC repo (which I am not in favour of but it is at this point too late) - access to this repo (read access included) must first thing be mirrored to people holding IAM roles able to access the secretsmanager:* permission via a role in Production (e.g. ExLabs colleagues etc.).

